PR #154, opened 2026-07-03
Verify microphone mute took effect; fail loudly on failure (#151)

jfbosch:
## What this fixes

Closes #151

The microphone mute control could tell the user they were muted while the
mic stayed live. Pressing mute played the mute sound and showed the muted
graphic, yet other people could still hear the user; only restarting the
app restored correct behaviour.

Two things caused the false confidence:

- `ToggleMute()` wrote the mute flag to each capture device inside a
  `try { } catch { }` that swallowed every error, returned `void`, and
  never read the value back — so a failed write (typically a stale
  CoreAudio COM proxy on a cached device) vanished silently.
- The beep and graphic were driven by `IsMuted`, which read the mute flag
  off a *different* device instance (the selected microphone) than the one
  the write targeted. The UI could therefore report "muted" from a stale
  flag while the live device was never muted.

## How it works now

A small, testable seam isolates the mute logic from the CoreAudio COM
types so the verify-and-retry behaviour can be unit-tested without audio
hardware:

- **`IMuteEndpoint`** — one device we can set and read the mute flag on;
  both operations may throw.
- **`IMuteEndpointProvider`** — hands out the endpoints and can
  re-enumerate fresh device references.
- **`MuteStateController`** — writes the desired state to every device,
  reads it back to confirm, and retries **once** against freshly-acquired
  device references when a write throws or the read-back disagrees. This
  reproduces automatically what a manual app restart used to do. The
  tracked mute state advances only on a confirmed success.
- **`MmDeviceMuteEndpoint`** — the thin CoreAudio adapter (mirrors the
  existing `CoreAudioCaptureLevelController` pattern).

`AudioDeviceManager` now delegates to the controller, `ToggleMute()`
returns a `MuteToggleResult`, and `IsMuted` reflects the aggregate state
that was actually written and verified.

In the UI, a confirmed toggle behaves exactly as before (mute/unmute beep,
graphic, status). On failure the app plays the failure beep, leaves the
graphic unchanged (so it never shows "muted" for a mute that did not take
effect), and shows an error in the status bar. This matters because a
blind user relies on the beep and graphic as the only confirmation that
muting worked — a false "you are muted" is worse than no signal.

## Acceptance criteria addressed

- [x] `ToggleMute()` writes the mute state, then reads `.Mute` back on each
      device; a thrown exception or read-back mismatch counts as failure.
- [x] `ToggleMute()` reports success/failure to the caller (returns
      `MuteToggleResult`) instead of `void`.
- [x] On confirmed success: mute/unmute beep, muted/live graphic, normal
      status message — as before.
- [x] On failure: failure beep, no muted graphic, error surfaced in the
      status bar.
- [x] Re-acquire fresh device references and retry once on a stale-proxy
      failure before declaring failure.
- [x] Reported mute state reflects the devices actually written and
      verified, not a stale read of one unrelated device instance.

## Test plan

`dotnet test --configuration Release` — all 509 tests pass, including 9
new `MuteStateControllerTests` covering:

- healthy devices confirm and advance state without a refresh;
- a second toggle unmutes;
- read-back mismatch triggers exactly one retry with fresh references, and
  the fresh reference is what actually mutes;
- a thrown write recovers via the retry;
- a partial failure (one bad device) fails the first attempt then retries;
- persistent failure returns failure, retries exactly once (no loop), and
  never advances the tracked state;
- a failed unmute while muted keeps the muted state (never falsely reports
  live);
- an empty device set is a failure (mute cannot be confirmed).

Manual verification of the live audio path (real hardware, stale-proxy
recovery) is left to a human, since it cannot be exercised in the unit
test environment.

## Notes

Related to but distinct from #150 (microphones missing from the device
list). This PR does not re-resolve the selected microphone during the
retry — waveform capture uses an integer device index, and device-list
refresh is #150's scope.

### DIFF
--- a/Mutation.Tests/MuteStateControllerTests.cs
+++ b/Mutation.Tests/MuteStateControllerTests.cs
@@ -1,0 +1,202 @@
+using System;
+using System.Collections.Generic;
+using Mutation.Ui.Core;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class MuteStateControllerTests
+{
+	// Stands in for a real audio endpoint without any COM hardware. It can be
+	// made to throw (a stale proxy) or to silently drop the write (read-back
+	// mismatch) so the verify-and-retry logic can be exercised deterministically.
+	private sealed class FakeMuteEndpoint : IMuteEndpoint
+	{
+		private bool _mute;
+
+		public bool ThrowOnSet { get; set; }
+		public bool ThrowOnGet { get; set; }
+		// When true, the write is accepted but the read-back reports the opposite
+		// state — i.e. the mute did not actually take effect.
+		public bool ReadBackMismatch { get; set; }
+		public int SetCount { get; private set; }
+		public bool CurrentMute => _mute;
+
+		public FakeMuteEndpoint(bool initial = false) => _mute = initial;
+
+		public void SetMute(bool mute)
+		{
+			SetCount++;
+			if (ThrowOnSet)
+				throw new InvalidOperationException("stale proxy");
+			_mute = mute;
+		}
+
+		public bool GetMute()
+		{
+			if (ThrowOnGet)
+				throw new InvalidOperationException("stale proxy");
+			return ReadBackMismatch ? !_mute : _mute;
+		}
+	}
+
+	private sealed class FakeProvider : IMuteEndpointProvider
+	{
+		private readonly IReadOnlyList<IMuteEndpoint> _initial;
+		private readonly IReadOnlyList<IMuteEndpoint> _refreshed;
+
+		public int RefreshCount { get; private set; }
+
+		public FakeProvider(IReadOnlyList<IMuteEndpoint> initial, IReadOnlyList<IMuteEndpoint>? refreshed = null)
+		{
+			_initial = initial;
+			_refreshed = refreshed ?? initial;
+		}
+
+		public IReadOnlyList<IMuteEndpoint> GetEndpoints() => _initial;
+
+		public IReadOnlyList<IMuteEndpoint> RefreshEndpoints()
+		{
+			RefreshCount++;
+			return _refreshed;
+		}
+	}
+
+	[Fact]
+	public void Toggle_AllEndpointsHealthy_ConfirmsAndAdvancesState_WithoutRefresh()
+	{
+		var a = new FakeMuteEndpoint();
+		var b = new FakeMuteEndpoint();
+		var provider = new FakeProvider(new IMuteEndpoint[] { a, b });
+		var controller = new MuteStateController(provider);
+
+		var result = controller.Toggle();
+
+		Assert.True(result.Success);
+		Assert.True(result.IsMuted);
+		Assert.True(controller.IsMuted);
+		Assert.True(a.CurrentMute);
+		Assert.True(b.CurrentMute);
+		Assert.Equal(0, provider.RefreshCount);
+	}
+
+	[Fact]
+	public void Toggle_Twice_UnmutesBackToLive()
+	{
+		var a = new FakeMuteEndpoint();
+		var provider = new FakeProvider(new IMuteEndpoint[] { a });
+		var controller = new MuteStateController(provider);
+
+		controller.Toggle();
+		var result = controller.Toggle();
+
+		Assert.True(result.Success);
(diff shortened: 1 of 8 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR fixes a privacy-critical bug where the mute control could report "muted" while the microphone stayed live. It introduces a small, testable seam (`IMuteEndpoint` / `IMuteEndpointProvider` / `MuteStateController`) that writes the desired mute state to every capture device, reads it back to confirm, and retries once against freshly-enumerated device references — reproducing automatically what a manual app restart used to do. The implementation is correct, minimal, well-layered, and thoroughly unit-tested.

### Non-compliance
None. The solution builds and the full suite is green (`dotnet test --configuration Release` → 509 passed, 0 failed, including the 9 new `MuteStateControllerTests`). Screen-reader accessibility is preserved: `UpdateMicrophoneToggleVisuals()` drives `AutomationProperties.SetName` on both `BtnToggleMic` and `MicStatusIcon` off the *confirmed* state, and a failed toggle plays `BeepType.Failure` and surfaces an `InfoBarSeverity.Error` status rather than a false "muted" signal. No hardcoded defaults hide behavior; no secrets committed.

### Story alignment
Matches issue #151 exactly. All six acceptance criteria are met:
- `ToggleMute()` writes then reads `.Mute` back on each device; a throw or read-back mismatch is a failure (`MuteStateController.TryApplyAndVerify`, `Mutation.Ui/Core/MuteStateController.cs:58`).
- `ToggleMute()` now returns `MuteToggleResult` instead of `void` (`AudioDeviceManager.cs:105`).
- Confirmed success keeps the prior beep/graphic/status behavior; failure plays the failure beep, leaves the graphic unchanged, and shows an error (`MainWindow.xaml.cs:503-514`).
- Fresh device references are re-acquired and the write retried exactly once on failure (`MuteStateController.cs:43`).
- Reported state reflects the devices actually written and verified, not a stale read of one unrelated instance (`AudioDeviceManager.IsMuted` now delegates to the controller, `AudioDeviceManager.cs:37`).

### Correctness
Traced the toggle path with concrete values. `Toggle()` computes `desired = !_isMuted`, applies-and-verifies against the current endpoints, and only on failure re-enumerates and retries once; `_isMuted` advances solely on a confirmed success, so a failed toggle never falsely reports "muted"/"live". An empty endpoint set is correctly treated as failure (a mute cannot be confirmed to have taken effect). Nullability is handled throughout — constructor guards, `BuildEndpoints()` filters nulls, and `MmDeviceMuteEndpoint` throws a clear `InvalidOperationException` when a device exposes no `AudioEndpointVolume`.

One inherent edge case worth noting (not blocking, no fix warranted): with *multiple* capture devices, a partially-failed **unmute** can leave the tracked state at "muted" while one device is already live. This is unavoidable without a rollback that could itself fail, and it is well mitigated — the failure path plays the failure beep and shows an explicit "The microphone may still be live — please try again" error, so the user (including a blind user relying on the beep) is warned rather than given false confidence. This is the safe direction for a privacy control.

### Tests
Excellent coverage. The 9 new tests exercise every branch via hardware-free fakes: healthy confirm-without-refresh, second-toggle-unmutes, read-back-mismatch → one retry with fresh refs (asserting the fresh reference is what actually mutes), thrown-write recovery, partial failure, persistent failure (returns failure, retries exactly once — no loop, state never advances), failed-unmute-keeps-muted, empty device set, and null-provider guard. The only untested path is the live CoreAudio hardware, which is correctly documented as manual verification.

### Regressions
The `ToggleMute()` signature changed from `void` to `MuteToggleResult`. The only call site is `BtnToggleMic_Click` (updated). The startup beep at `MainWindow.xaml.cs:224` reads `IsMuted`, which is still valid and now reflects the verified aggregate state. No other consumers exist. `RefreshCaptureDevices()` is untouched.

### Minimality
Every changed line serves the fix. No unrelated refactors, formatting churn, or bundled work. The new interfaces follow the existing `CoreAudioCaptureLevelController` adapter pattern, and UI→Core layering is respected (no UI types leak into Core).

### Fixes applied

No issues remaining.

---
Reviewed at add528b7065fd299a684f788909abc02241dc2a7
🤖 Reviewed with Claude Code

PR #155, opened 2026-07-03
Re-enumerate capture devices on hot-plug so the mute hotkey covers new mics

jfbosch:
## What and why

The microphone mute/unmute hotkey is meant to silence **every** capture
device — a privacy guarantee a blind user relies on. But the list of devices
it acted on was built only once, at startup, so a microphone plugged in after
the app launched was never muted. Pressing mute left that mic live, with
nothing to signal the miss. This fixes that silent failure.

## What changed

The app now listens for OS audio device-change notifications. When a device is
added, removed, or changes state, it re-enumerates capture devices and re-applies
the current mute state, so:

- The hotkey always acts on the current set of microphones, including any
  connected mid-session.
- A microphone connected while the app is muted is muted too — state stays
  consistent across all mics.

New pieces, kept small and layered so the logic stays testable:

- **`ICaptureDeviceChangeNotifier`** — an interface over "the device set
  changed", so `AudioDeviceManager` reacts to hot-plugs without a direct
  dependency on the CoreAudio COM API.
- **`MMDeviceCaptureDeviceChangeNotifier`** — a thin CoreAudio wrapper around
  `MMNotificationClient` that collapses the per-device events into one
  `CaptureDevicesChanged` signal.
- **`MuteStateController.SynchronizeDevices()`** — re-applies the current
  confirmed mute state to the current endpoint set. Best-effort: it never
  changes the tracked state (a real toggle owns state transitions and their
  verification). This is where "a new mic adopts the current state" lives, and
  it is fully unit-tested.
- **`AudioDeviceManager`** — injects the notifier and, on a change, refreshes
  and synchronizes under a lock so the background notification thread never
  races the UI-thread mute toggle over the shared device list.

The issue's secondary note — that `IsMuted` was derived from a single device —
was already resolved by an earlier refactor: `IsMuted` reflects the aggregate
confirmed state via `MuteStateController`, so no further reconciliation was
needed.

## Test plan

- `dotnet test --configuration Release` — all 514 tests pass, including five new
  `MuteStateController.SynchronizeDevices` tests:
  - a newly added endpoint adopts the muted state while the app is muted;
  - a mic present when the app started muted is muted without a toggle;
  - when unmuted, sync leaves new devices live and does not change tracked state;
  - a partial failure reports failure and never advances/flips tracked state;
  - an empty endpoint set returns false without throwing.
- Manual (hardware): with the app muted, plug in a second microphone and confirm
  it comes up muted; with the app live, confirm a newly plugged mic stays live;
  press the mute hotkey after hot-plugging and confirm all mics toggle together.

## Notes

Any device change (render devices included) triggers a re-enumeration. Capture
re-enumeration is cheap and idempotent, so this is a deliberate simplicity
tradeoff over filtering each event by data-flow. `AudioDeviceManager` remains a
COM adapter that is exercised through hardware rather than unit tests, matching
the existing design; the new behavioural guarantee is tested at the
`MuteStateController` seam.

Closes #150

### DIFF
--- a/Mutation.Tests/MuteStateControllerTests.cs
+++ b/Mutation.Tests/MuteStateControllerTests.cs
@@ -199,4 +199,101 @@ public class MuteStateControllerTests
 	{
 		Assert.Throws<ArgumentNullException>(() => new MuteStateController(null!));
 	}
+
+	// A provider whose current endpoint set can grow, standing in for a
+	// microphone hot-plugged after the app started. GetEndpoints() reflects the
+	// latest set, exactly as AudioDeviceManager does after re-enumerating.
+	private sealed class MutableProvider : IMuteEndpointProvider
+	{
+		private IReadOnlyList<IMuteEndpoint> _current;
+
+		public int RefreshCount { get; private set; }
+
+		public MutableProvider(IReadOnlyList<IMuteEndpoint> initial) => _current = initial;
+
+		public void SetEndpoints(IReadOnlyList<IMuteEndpoint> endpoints) => _current = endpoints;
+
+		public IReadOnlyList<IMuteEndpoint> GetEndpoints() => _current;
+
+		public IReadOnlyList<IMuteEndpoint> RefreshEndpoints()
+		{
+			RefreshCount++;
+			return _current;
+		}
+	}
+
+	[Fact]
+	public void SynchronizeDevices_WhenMuted_AppliesMutedToNewlyAddedEndpoint()
+	{
+		var existing = new FakeMuteEndpoint();
+		var provider = new MutableProvider(new IMuteEndpoint[] { existing });
+		var controller = new MuteStateController(provider);
+
+		controller.Toggle(); // now muted; the existing mic is muted
+		Assert.True(controller.IsMuted);
+
+		// A mic is plugged in: it joins the current set, initially live.
+		var added = new FakeMuteEndpoint();
+		provider.SetEndpoints(new IMuteEndpoint[] { existing, added });
+
+		bool synced = controller.SynchronizeDevices();
+
+		Assert.True(synced);
+		Assert.True(added.CurrentMute);       // the new mic adopted the muted state
+		Assert.True(existing.CurrentMute);    // the existing mic stays muted
+		Assert.True(controller.IsMuted);      // tracked state is unchanged
+	}
+
+	[Fact]
+	public void SynchronizeDevices_WhenStartedMuted_MutesNewMic_WithoutAToggle()
+	{
+		var added = new FakeMuteEndpoint();
+		var provider = new MutableProvider(new IMuteEndpoint[] { added });
+		var controller = new MuteStateController(provider, initialMuted: true);
+
+		bool synced = controller.SynchronizeDevices();
+
+		Assert.True(synced);
+		Assert.True(added.CurrentMute);
+	}
+
+	[Fact]
+	public void SynchronizeDevices_WhenUnmuted_LeavesNewMicLive_AndDoesNotChangeState()
+	{
+		var added = new FakeMuteEndpoint(initial: true); // arrives muted at the OS level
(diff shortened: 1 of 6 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR fixes a silent-failure bug where the mute hotkey missed microphones hot-plugged after startup. It adds OS device-change notifications that re-enumerate capture devices and re-apply the current mute state, so every mic — including newly connected ones — stays consistent with the app's mute state. The implementation is correct, well-layered, and thoroughly tested.

### Non-compliance
None. Build is clean (`dotnet build --configuration Release`, 0 warnings / 0 errors) and the full suite passes (`dotnet test --configuration Release`, 514/514). No accessibility surface changes and no hardcoded-default gate applies (this is a correctness fix, not a hidden setting).

### Story alignment
Fully implements issue #150. All three acceptance criteria are met: capture devices are re-enumerated on OS device change (`AudioDeviceManager.OnCaptureDevicesChanged`), a newly connected mic adopts the current mute state (`MuteStateController.SynchronizeDevices`), and OS notifications via `IMMNotificationClient`/CoreAudio drive the refresh rather than startup-only enumeration. The issue's secondary `IsMuted`-from-one-device concern was already resolved by the earlier `MuteStateController` refactor, correctly noted in the PR body — `IsMuted` now reflects the aggregate confirmed state.

### Correctness
- Threading is handled carefully. `AudioDeviceManager` serializes the background notification thread against the UI-thread `ToggleMute` under a single reentrant `_sync` lock, and `CaptureDevices`/`GetEndpoints` return snapshots so an enumerator can't be swapped mid-iteration (`AudioDeviceManager.cs:41`, `:54`, `:139`). `_isMuted` in `MuteStateController` is only ever read/written under that lock (`Toggle` and `SynchronizeDevices` are both called inside it), so there's no torn state.
- Construction ordering is safe: `RefreshCaptureDevices()` and `_muteState` are set up before subscribing to `CaptureDevicesChanged`, so a notification arriving during construction can't reach a half-built object (`AudioDeviceManager.cs:30-35`).
- `SynchronizeDevices()` is correctly best-effort: it re-applies the tracked state without ever advancing it, so a real toggle remains the sole owner of state transitions (`MuteStateController.cs:52-62`).
- DI wiring injects the same singleton `MMDeviceEnumerator` into both the manager and the notifier, so notifications are registered on the enumerator the app actually uses (`App.xaml.cs:193`).
- Design note (not a defect): the CoreAudio notification callback re-enumerates and mutes on the callback thread while briefly holding `_sync`. This matches the common NAudio/CoreAudio pattern and the lock is never held across a call back into the audio engine, so there is no lock-ordering cycle. The re-enumeration is deliberately unfiltered (render devices included) per the documented cheap-and-idempotent tradeoff. Manual hardware testing in the PR confirms the behaviour.

### Tests
Strong. Five new `SynchronizeDevices` tests cover the meaningful paths: a new endpoint adopting the muted state, a mic present when the app started muted, the unmuted case leaving new devices live without changing tracked state, a partial failure reporting failure without advancing state, and the empty-set case. The `MuteStateController` seam is the right place to test the new guarantee deterministically without COM hardware; `AudioDeviceManager` remains a thin COM adapter exercised through hardware, consistent with the existing design.

### Regressions
None found. `ToggleMute` retains its verify-and-retry semantics; the only change is that it now runs under the shared lock. The `IMuteEndpointProvider` refresh path is unchanged for existing callers. No existing consumers of `AudioDeviceManager` are broken by the added constructor parameter (resolved via DI).

### Minimality
Every change is necessary and scoped to the fix: the new notifier interface + CoreAudio implementation, the `SynchronizeDevices` method, the manager wiring, and the DI registration. No unrelated refactors or formatting churn.

### Fixes applied

No issues remaining.

---
Reviewed at d1ea22ce536982d851530f0b864e63040713b6bf
🤖 Reviewed with Claude Code

PR #157, opened 2026-07-03
Verify pinned mic level took effect and fail loudly on failure (#153)

### LINKED ISSUE #153: [BUG] Pinned mic input level can silently fail to apply — no verification, no failure feedback
## Context

The app can "pin" a preferred microphone input level (gain) and re-asserts it right before recording so that a level another app changed is corrected back to the user's choice. But if that re-assertion **fails**, the failure is silent — the app records anyway, at whatever gain the device was left on, with no beep, status, or any other indication that the pin did not take.

This is the same silent-failure class as the mute bug (#151), but for input **gain** instead of mute. It is not covered by #151 (mute) or #152 (foreground slider refresh).

**How the failure is swallowed, end to end:**

1. `CoreAudioCaptureLevelController.SetLevelScalar()` writes `MasterVolumeLevelScalar` inside a `try/catch` that discards every exception — `Mutation.Ui/Core/CoreAudioCaptureLevelController.cs:41` (catch at `:53`).

```csharp
public void SetLevelScalar(float scalar)
{
    var volume = _activeDevice()?.AudioEndpointVolume;
    if (volume is null)
        return;

    float clamped = ...;
    try { volume.MasterVolumeLevelScalar = clamped; }
    catch { /* skip silently */ }        // failure discarded, no read-back
}
```

2. `MicrophoneLevelPinService.WriteLevel()` returns a bool signalling whether it actually wrote — `Mutation.Ui/Core/MicrophoneLevelPinService.cs:56`. But that bool only reflects "we issued a write / it wasn't a redundant no-op", not "the hardware level is now the target" — there is no read-back verification.

3. The record-start caller **ignores the return value** entirely and proceeds straight into recording — `Mutation.Ui/Core/AudioSessionManager.cs:165`:

```csharp
// Re-assert the pinned capture level right before recording ...
_levelPinService.ReassertPinnedLevel(_settings.AudioSettings?.PinnedCaptureLevel);
// ... recording starts regardless of whether the level actually took
```

The `ApplyLevel()` return is likewise ignored at the slider/toggle handlers (`Mutation.Ui/MainWindow.xaml.cs:1206` and `:1224`).

## Why this happens in practice

The likely failure is a stale/disconnected CoreAudio COM proxy on the cached `MMDevice` (the same condition behind #151 and the hot-plug gap in #150). When the proxy is stale, `SetLevelScalar` throws, the exception is swallowed, and the app records at the wrong gain.

## Impact

Medium. The recording is captured at an unintended gain (too quiet — poor transcription — or too hot/clipping), and the user gets no signal that the pin was skipped. For a blind user who relies on the pin to guarantee a consistent capture level, a silent miss means a bad recording discovered only after the fact.

## Where

- `Mutation.Ui/Core/CoreAudioCaptureLevelController.cs:41` — `SetLevelScalar()` swallows the write failure, no read-back.
- `Mutation.Ui/Core/MicrophoneLevelPinService.cs:56` — `WriteLevel()` return reflects "write issued", not "level verified".
- `Mutation.Ui/Core/AudioSessionManager.cs:165` — record-start re-assert ignores the result and records regardless.
- `Mutation.Ui/MainWindow.xaml.cs:1206`, `:1224` — `ApplyLevel()` result ignored on toggle/slider.
- `CognitiveSupport/BeepPlayer.cs:5` — `BeepType.Failure` already exists for the failure signal.

## Requirements (acceptance criteria)

- `SetLevelScalar()` / the pin service **verify the write** by reading the level back and confirming it matches the target within the existing epsilon (`MicrophoneLevelPinService.LevelEpsilon`); a thrown exception or a read-back mismatch counts as failure.
- The pin write reports success/failure to callers instead of the current "did we issue a write" bool.
- **Re-acquire and retry once on a stale-proxy failure:** when the write fails, re-enumerate / re-resolve the active device to obtain fresh `MMDevice` / `AudioEndpointVolume` references and retry once before declaring failure. (Shared machinery with #151 — build them consistently.)
- **At record start:** if the pinned level cannot be applied and verified, surface it — do not silently record. At minimum play `BeepType.Failure` and show an error/warning in the status bar so the user knows the capture level is not the pinned value. (Decide whether to still proceed with recording or abort; the story requires the user be *told*, not silently proceed.)
- **On live slider drag / pin toggle:** if applying the level fails, reflect it to the user (failure beep and/or status) rather than showing the slider at a value the hardware never accepted.
- Preserve the existing graceful behaviour for genuinely hardware-fixed devices (`IsLevelControlSupported == false`): those are legitimately uncontrollable and should not raise a failure signal on every attempt — distinguish "unsupported" from "supported but the write failed".

## Notes

- Shares the "re-acquire fresh device references + verify + retry once" pattern with #151 (mute silent failure). Coordinate the two so the device-refresh/verify helper is written once and reused. Also related to #150 (device re-enumeration).
- `AudioSessionManager` and `OcrManager` already use `BeepType.Failure`; follow the same pattern.

## Blocks current story?

No. Filed for pickup.

jfbosch:
Closes #153

## What this fixes

The app lets you "pin" a preferred microphone input level (gain) and re-asserts it just before recording, so a level another app changed is corrected back to your choice. But when that re-assertion failed, the failure was silent: the app recorded anyway, at whatever gain the device was left on, with no beep, status, or any other sign that the pin never took. For a blind user relying on the pin to guarantee a consistent capture level, a silent miss meant a bad recording discovered only after the fact.

This is the input-gain counterpart of the mute silent-failure fixed in #151, and it deliberately reuses the same recovery shape so the two behave consistently.

## How it works now

The old design wrote the level inside a `try/catch` that discarded every exception and reported only "did we issue a write" — never "the level is actually the target now." That is replaced with three small pieces that mirror the mute fix:

- **A thin endpoint** (`MmDeviceCaptureLevelEndpoint`) that reads and writes one device's level and *throws* on failure instead of swallowing it.
- **A provider** (implemented by `AudioDeviceManager`) that hands out the active microphone's endpoint and can re-resolve the microphone to a fresh set of CoreAudio references — the same recovery a manual app restart used to perform by hand.
- **Verify-and-retry orchestration** in `MicrophoneLevelPinService`: it writes the level, reads it back, and only reports success if the read-back matches the target within the existing epsilon. If the write throws or the read-back does not match, it re-acquires fresh device references and retries exactly once before declaring failure.

The service now returns a typed result — `Unsupported`, `Unchanged`, `Applied`, or `Failed` — so callers can tell a genuinely uncontrollable (hardware-fixed) device apart from a controllable device whose write actually failed.

### What the user now sees

- **At record start:** if the pinned level cannot be applied and verified, a failure beep plays (before capture begins, so it is not recorded) and a status message says the level could not be applied and recording is at the current level. Recording still proceeds, so no audio is lost — but you are told.
- **On the pin toggle and live slider drag:** a failed write plays the failure beep and shows an error, rather than leaving the slider at a value the hardware never accepted.
- **Hardware-fixed devices** (no controllable level) stay graceful: they resolve to `Unsupported` and never raise a false failure signal.

## Acceptance criteria

| Criterion | Where |
| --- | --- |
| Verify the write by read-back within `LevelEpsilon`; exception or mismatch = failure | `MicrophoneLevelPinService.TryApplyAndVerify` |
| Report success/failure instead of "did we issue a write" | `CaptureLevelResult` / `CaptureLevelOutcome` |
| Re-acquire fresh references and retry once on stale-proxy failure | `MicrophoneLevelPinService.WriteLevel` + `AudioDeviceManager.RefreshEndpoint`/`RefreshActiveMicrophone` |
| Record start surfaces failure (beep + status), does not silently record | `AudioSessionManager.StartStopRecordingAsync` |
| Slider/toggle reflect failure to the user | `MainWindow.ReportIfLevelWriteFailed` |
| Preserve graceful behaviour for hardware-fixed devices; distinguish unsupported from failed write | `Unsupported` outcome + endpoint `IsLevelControlSupported` |

## Test plan

- `dotnet test --configuration Release` — all 522 tests pass.
- `MicrophoneLevelPinServiceTests` was rewritten against the new provider/endpoint and now covers: read-back verification, retry-on-fresh-endpoint success, failure when both attempts throw, failure on a read-back mismatch after retry, and the graceful `Unsupported` path — none of which the old "write issued" bool could express.

The CoreAudio hardware paths (`MmDeviceCaptureLevelEndpoint`, `AudioDeviceManager` device re-resolution) are not unit-tested, matching the existing convention from #151 where the equivalent `MmDeviceMuteEndpoint` was also left to manual/hardware verification. Manual check to run on a machine with a controllable mic: pin a level, change the OS level from another app, start recording, and confirm the level is corrected; then exercise a stale-device scenario to confirm the failure beep and status message.


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR replaces the silent-failure input-gain path with a verify-and-retry design that mirrors the mute fix from #151: writes are confirmed by read-back within `LevelEpsilon`, a stale-proxy failure triggers one re-acquire-and-retry, and a typed `CaptureLevelResult` lets callers distinguish a genuinely uncontrollable device (`Unsupported`) from a write that actually failed (`Failed`). It does what the issue asks, correctly.

### Non-compliance
None. `dotnet test --configuration Release` builds clean and passes 522/522. No screen-reader regression (the failure path adds an audible beep plus an `InfoBarSeverity.Error` status message — more feedback, not less). New behaviour is surfaced through existing UI controls, no hidden defaults. No secrets.

### Story alignment
Every acceptance criterion is met: read-back verification within epsilon (`MicrophoneLevelPinService.TryApplyAndVerify`), success/failure reporting via `CaptureLevelResult`/`CaptureLevelOutcome`, re-acquire-fresh-references-and-retry-once (`WriteLevel` + `AudioDeviceManager.RefreshEndpoint`/`RefreshActiveMicrophone`), record-start beep+status without silently recording (`AudioSessionManager.StartStopRecordingAsync:170`), slider/toggle failure feedback (`MainWindow.ReportIfLevelWriteFailed`), and the `Unsupported` outcome preserving graceful behaviour for hardware-fixed devices.

### Correctness
- Record-start plays `BeepType.Failure` **before** `StartRecordingAsync`, so the failure beep is not captured into the recording — verified at `AudioSessionManager.cs:171-179`.
- Null-safe throughout: `MmDeviceCaptureLevelEndpoint` accepts a null device (no mic selected), `IsLevelControlSupported` returns false, and `WriteLevel` gates on it and returns `Unsupported` before any read/write, so there is no dereference of a missing device.
- The redundant-write skip reads through `TryReadScalar`, which swallows a stale-proxy read exception to `null` — so an unreadable current level correctly falls through to an attempted write rather than a false "unchanged".
- Retry ordering is right: the fresh endpoint is obtained from `RefreshEndpoint()` (which re-points `_microphone`) only after the first verify fails, and `RefreshActiveMicrophone` is a safe no-op when the previously-selected device is gone.

Non-blocking observations, all within spec and not worth changing:
- The startup / mic-change re-assert at `MainWindow.xaml.cs:1236` ignores the returned result and stays silent on failure. That is outside the issue's explicit AC scope (record start, slider drag, pin toggle), and beeping during initialization/device-switch would arguably be noisier than helpful — acceptable as-is.
- Read-back verification against a 1-unit epsilon could, in principle, report `Failed` on hardware that quantizes the endpoint scalar to coarse steps larger than 1%. This is exactly the epsilon the story mandated ("within the existing epsilon"), so it is by design; flagging only as future awareness.

### Tests
Strong. `MicrophoneLevelPinServiceTests` was rewritten against the new provider/endpoint and covers the outcomes the old "write issued" bool could not express: read-back verification, retry-on-fresh-endpoint success, failure when both attempts throw, failure on read-back mismatch after retry, and the graceful `Unsupported` path. The CoreAudio hardware endpoints are left to manual verification, matching the #151 convention.

### Regressions
None found. The deleted `ICaptureLevelController`/`CoreAudioCaptureLevelController` have no remaining references; DI now supplies `AudioDeviceManager` as the `ICaptureLevelEndpointProvider`. All `ReadCurrentLevel`/`ApplyLevel`/`ReassertPinnedLevel` call sites were updated to the new return type.

### Minimality
Tight. Every change serves the fix — new endpoint/provider abstractions, the typed result, verify-and-retry orchestration, and the two UI feedback points. No unrelated refactors. New files follow the tabs convention.

### Fixes applied

No issues remaining.

---
Reviewed at 7d0b53ad05879dc450a63be8387f414f53741aed
🤖 Reviewed with Claude Code

PR #182, opened 2026-07-03
Confirm before deleting an LLM prompt (#163)

jfbosch:
## What and why

Deleting a prompt from the LLM prompt library used to remove it and write the
change to disk immediately — with no confirmation, no screen-reader
announcement, no beep, and no undo. Because each row's **Delete** button sits
right beside **Edit** and **Run**, a single stray Enter while tabbing through
the list destroyed a carefully built prompt (name, model, hotkey, and full
instruction text) silently. A screen-reader user would only discover the loss
later, when the prompt's hotkey stopped working.

## What changed

- **Confirmation before removal.** Deleting now opens a confirmation dialog
  ("Delete prompt 'X'? This cannot be undone."). Its default button is
  **Cancel**, so a stray Enter dismisses the dialog instead of deleting.
- **Spoken outcome.** After a prompt is actually removed, the status InfoBar
  announces the result and a success beep plays. Cancelling is announced too,
  so the screen-reader user always hears what happened.
- **Accurate reporting.** `DeletePrompt` now reports whether it actually
  removed a prompt, so the announcement never claims a deletion that did not
  happen. The confirmation path also covers the case where another dialog is
  already open — in that case nothing is deleted.
- **No new silent failure.** The handler is an `async void`; if persisting the
  removal fails (e.g. the settings file cannot be written), the failure is
  announced with an error status and a failure beep rather than being
  swallowed by the global crash handler.
- **Testable wording.** The user-facing phrasing is extracted into a small,
  UI-free `PromptDeletionMessages` helper so the accessible messages are
  centralized and unit-tested. (The prompt-library controller holds a WinUI
  `ListView` and cannot be built in a plain unit test, so this helper is the
  testable seam.)

## Files

- `Mutation.Ui/Services/PromptDeletionMessages.cs` — new pure helper for the
  confirmation question and the deleted/cancelled announcements, with a
  neutral fallback for an unnamed prompt.
- `Mutation.Ui/Services/PromptLibraryController.cs` — `DeletePrompt` returns
  whether a prompt was removed.
- `Mutation.Ui/MainWindow.xaml.cs` — the Delete handler now confirms, deletes
  on confirm, and announces the outcome with a beep.
- `Mutation.Tests/PromptDeletionMessagesTests.cs` — tests for the helper.

## Test plan

- `dotnet test --configuration Release` — all 536 tests pass, including 14 new
  cases for `PromptDeletionMessages` (named prompt, whitespace-trimmed name,
  and blank/null-name fallback across the confirmation, both announcements,
  and the failure message with and without a reason).
- Manual, with a screen reader: tab to a prompt's **Delete** button and press
  Enter. Confirm the dialog opens, Cancel is the focused default, pressing
  Enter/Escape cancels without deleting (and the cancellation is announced),
  and choosing **Delete** removes the prompt with a spoken confirmation and a
  beep.

Closes #163

### DIFF
--- a/Mutation.Tests/PromptDeletionMessagesTests.cs
+++ b/Mutation.Tests/PromptDeletionMessagesTests.cs
@@ -1,0 +1,58 @@
+using Mutation.Ui.Services;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class PromptDeletionMessagesTests
+{
+	[Fact]
+	public void BuildConfirmation_NamedPrompt_QuotesNameAndWarnsIrreversible()
+	{
+		string message = PromptDeletionMessages.BuildConfirmation("Summarize");
+
+		Assert.Equal("Delete prompt 'Summarize'? This cannot be undone.", message);
+	}
+
+	[Fact]
+	public void BuildConfirmation_TrimsSurroundingWhitespaceInName()
+	{
+		string message = PromptDeletionMessages.BuildConfirmation("  Summarize  ");
+
+		Assert.Equal("Delete prompt 'Summarize'? This cannot be undone.", message);
+	}
+
+	[Theory]
+	[InlineData(null)]
+	[InlineData("")]
+	[InlineData("   ")]
+	public void BuildConfirmation_BlankName_FallsBackToNeutralLabel(string? name)
+	{
+		string message = PromptDeletionMessages.BuildConfirmation(name);
+
+		Assert.Equal("Delete this prompt? This cannot be undone.", message);
+	}
+
+	[Fact]
+	public void BuildDeleted_NamedPrompt_ReportsRemoval()
+	{
+		Assert.Equal("Deleted prompt 'Summarize'.", PromptDeletionMessages.BuildDeleted("Summarize"));
+	}
+
+	[Fact]
+	public void BuildDeleted_BlankName_FallsBackToNeutralLabel()
+	{
+		Assert.Equal("Deleted this prompt.", PromptDeletionMessages.BuildDeleted("  "));
+	}
+
+	[Fact]
+	public void BuildCancelled_NamedPrompt_ReportsCancellation()
+	{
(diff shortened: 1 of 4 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR adds a confirmation dialog before deleting an LLM prompt, with Cancel as the safe default button, and announces every outcome (deleted, cancelled, failed) via the status InfoBar plus a beep. It correctly implements what issue #163 asked for.

### Non-compliance
None. Solution builds and all 536 tests pass (`dotnet test --configuration Release`). The change improves screen-reader accessibility rather than regressing it: the dialog carries `AutomationProperties` name and help text, and every outcome is announced.

### Story alignment
Fully aligned with issue #163: confirmation ContentDialog with the exact suggested wording, outcome announced via `ShowStatus` plus a beep. Nothing out of scope.

### Correctness
The handler guards against a null controller, uses the existing `ShowDialogAsync` gate so a second open dialog cancels safely, and `PromptLibraryController.DeletePrompt` now returns whether a prompt was actually removed so the announcement is always accurate. The `async void` handler wraps the persistence call in try/catch and fails loudly with an error status and failure beep. One gap was found and fixed during review (see Fixes applied).

### Tests
14 new unit tests cover `PromptDeletionMessages` — named, whitespace-trimmed, and blank/null-name fallbacks across all four messages, and the failure message with and without a reason. The dialog flow itself lives in `MainWindow.xaml.cs`, which is not unit-testable; extracting the wording into a UI-free helper is a reasonable testable seam.

### Regressions
None. `BtnDeletePrompt_Click` is the only caller of `DeletePrompt`, and its signature change (void → bool) is safe.

### Minimality
All changes are necessary for the stated purpose. No unrelated refactors.

### Fixes applied
- 9b69fcb — When the user confirms but `DeletePrompt` returns false (the prompt was already gone), the handler was silent — the same silent-no-feedback pattern the issue fixes. It now announces the failure with a warning status and failure beep.

No issues remaining.

---
Reviewed at 9b69fcb9e79760080471a28d7363b693abbb8e0e
🤖 Reviewed with Claude Code

PR #183, opened 2026-07-03
Surface silent core and prompt hotkey registration failures (#162)

### LINKED ISSUE #162: [BUG] Core and prompt hotkey registration failures are silent
# Context

When registration of a core global hotkey (for example the dictation toggle) or a prompt hotkey fails — because another app owns the combination or the configured text cannot be parsed — the failure is only written to debug output. The user gets no dialog, no beep, no status message. Router hotkey failures are already surfaced properly, so core and prompt hotkeys lack an equivalent that already exists in the codebase.

# Evidence

- `Mutation.Ui/MainWindow.xaml.cs:426-430` — `TryRegister` catches all parse/register exceptions and only calls `Debug.WriteLine`.
- `Mutation.Ui/Services/HotkeyManager.cs:282-311` — `RegisterPromptHotkeys` catches and logs; it adds to `FailedRegistrations`, which nothing on the UI reads for these paths.
- Contrast `MainWindow.xaml.cs:2120-2125` — router hotkey failures are surfaced via `ShowRouterBindingFailuresAsync`.
- Related: `Mutation.Ui/Views/PromptEditorWindow.xaml.cs:92` stores the prompt hotkey as raw unvalidated text instead of using the validating `HotkeyEditor` control.

# Impact

The primary user drives this app almost entirely through global hotkeys. A collision or typo means the hotkey simply never fires, with zero feedback and no way to discover why.

# Suggested fix

Collect registration failures from `RegisterCoreHotkeys` and `RegisterPromptHotkeys` and surface them the same way router failures are surfaced (dialog on startup/settings-apply, plus the failure beep). Reuse `HotkeyEditor`/`HotkeyValidator` in `PromptEditorWindow`.

Found during comprehensive codebase audit. Does not block any current story.

jfbosch:
Closes #162

## Problem

Global-hotkey registration failures on the **core** and **prompt** paths
were only written to debug output. When a shortcut collided with another
app or the configured text could not be parsed, the hotkey simply never
fired — no dialog, no beep, no status message. For a user who drives the
app almost entirely by hotkey, a collision or a typo was invisible and
undiagnosable. Router-hotkey failures were already surfaced with a dialog,
so the core and prompt paths were the odd ones out.

The worst case was a **typo**: `MainWindow.TryRegister` caught the parse
exception and only logged it, so a mistyped core hotkey never even reached
the failure list the startup dialog reads from. On settings-apply, only
router failures were shown.

## What changed

The three registration paths (core, router, prompt) now report failures
through one shared shape and are surfaced the same way: a failure beep
plus a single accessible dialog listing each hotkey that could not be
bound and why.

- **`HotkeyManager`** — a new `HotkeyBindingFailure` record plus three
  pure, unit-testable helpers: `ClassifyConfiguredHotkey` (catches the
  previously-swallowed parse errors), `BuildFailureMessage` (renders the
  dialog text), and `ToBindingFailures` (maps router results into the
  shared shape). `TryRegisterCore` returns the failure reason instead of
  discarding it, and `RegisterPromptHotkeys` returns per-prompt failures
  labelled by prompt name.
- **`MainWindow`** — `RegisterCoreHotkeys` collects failures with a
  human-readable label for each action; `ShowHotkeyBindingFailuresAsync`
  plays the failure beep and shows one dialog for core, router, and prompt
  failures on settings-apply.
- **App startup** — collects and surfaces all hotkey failures the same
  way, and drops a redundant second router registration.
- **`PromptLibraryController`** — reports prompt-hotkey failures after a
  prompt is added or edited.
- **`PromptEditorWindow`** — the raw hotkey text box is replaced with the
  validating `HotkeyEditor` control, and an invalid hotkey is now rejected
  on save instead of being stored as unvalidated text.

## Accessibility

The dialog carries an automation name and help text so a screen reader
announces the full list of failures, and the failure beep gives an
immediate non-visual cue. Each failing hotkey is named by its action
(e.g. "Start/stop dictation") or prompt name, not just the raw key
string.

## Test plan

- `dotnet test --configuration Release` — all 538 tests pass, including 15
  new tests covering failure classification, message formatting, and the
  router-result mapping.
- Manual (device-dependent, not automated): configure a core hotkey that
  collides with another app or contains a typo → on startup and on
  settings-apply a dialog lists the failure and the failure beep plays.
  Add a prompt with an invalid hotkey in the prompt editor → save is
  blocked with an inline validation message.

## Notes

Registration itself uses the Win32 `RegisterHotKey` P/Invoke, which needs
a live window handle, so the collision path cannot be unit-tested. The new
tests target the pure classification and formatting logic (which is where
the silent parse-failure bug lived); the validation reuse in the prompt
editor is covered by the existing `HotkeyValidator` tests.

### DIFF
--- a/Mutation.Tests/HotkeyBindingFailureTests.cs
+++ b/Mutation.Tests/HotkeyBindingFailureTests.cs
@@ -1,0 +1,130 @@
+using CognitiveSupport;
+using Mutation.Ui.Services;
+
+namespace Mutation.Tests;
+
+public class HotkeyBindingFailureTests
+{
+	[Theory]
+	[InlineData("Ctrl+C")]
+	[InlineData("Ctrl+Shift+F5")]
+	[InlineData("Ctrl+Alt+S")]
+	public void ClassifyConfiguredHotkey_ValidHotkey_ReturnsNull(string input)
+	{
+		Assert.Null(HotkeyManager.ClassifyConfiguredHotkey("Some action", input, allowEmpty: true));
+	}
+
+	[Theory]
+	[InlineData("NotAKey")]
+	[InlineData("Ctrl+Shift")]
+	[InlineData("Ctrl+Bogus")]
+	public void ClassifyConfiguredHotkey_InvalidHotkey_ReturnsFailureWithReason(string input)
+	{
+		var failure = HotkeyManager.ClassifyConfiguredHotkey("Start/stop dictation", input, allowEmpty: true);
+
+		Assert.NotNull(failure);
+		Assert.Equal("Start/stop dictation", failure!.Description);
+		Assert.Equal(input.Trim(), failure.Hotkey);
+		Assert.False(string.IsNullOrWhiteSpace(failure.Reason));
+	}
+
+	[Theory]
+	[InlineData(null)]
+	[InlineData("")]
+	[InlineData("   ")]
+	public void ClassifyConfiguredHotkey_EmptyWhenAllowed_ReturnsNull(string? input)
+	{
+		Assert.Null(HotkeyManager.ClassifyConfiguredHotkey("Optional action", input, allowEmpty: true));
+	}
+
+	[Fact]
+	public void ClassifyConfiguredHotkey_EmptyWhenNotAllowed_ReturnsFailure()
+	{
+		var failure = HotkeyManager.ClassifyConfiguredHotkey("Required action", "", allowEmpty: false);
+
+		Assert.NotNull(failure);
+		Assert.Equal("Required action", failure!.Description);
+		Assert.Equal("Enter a hotkey.", failure.Reason);
+	}
+
+	[Fact]
+	public void BuildFailureMessage_EmptyList_ReturnsEmptyString()
+	{
+		Assert.Equal(string.Empty, HotkeyManager.BuildFailureMessage(new List<HotkeyManager.HotkeyBindingFailure>()));
+	}
+
+	[Fact]
+	public void BuildFailureMessage_WithHotkey_IncludesDescriptionHotkeyAndReason()
+	{
+		var failures = new List<HotkeyManager.HotkeyBindingFailure>
+		{
+			new("Start/stop dictation", "Ctrl+Alt+S", "The shortcut is already registered by another application."),
+		};
+
+		var message = HotkeyManager.BuildFailureMessage(failures);
+
+		Assert.Equal("• Start/stop dictation (Ctrl+Alt+S): The shortcut is already registered by another application.", message);
+	}
+
+	[Fact]
+	public void BuildFailureMessage_WithoutHotkey_OmitsParentheses()
+	{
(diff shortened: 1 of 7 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

The PR routes core, router, and prompt hotkey registration failures through one shared `HotkeyBindingFailure` shape and surfaces them with a failure beep plus a single accessible dialog, exactly as issue #162 asks. The logic is correct and well tested.

### Non-compliance
None. The solution builds and all 538 tests pass (`dotnet test --configuration Release`).

### Story alignment
All four evidence points in issue #162 are addressed: the swallowed parse exception in `MainWindow.TryRegister` now returns a failure via `HotkeyManager.TryRegisterCore`; prompt registration failures are returned and surfaced; core/prompt failures use the same dialog-plus-beep path router failures already had; and `PromptEditorWindow` now uses the validating `HotkeyEditor` control and rejects an invalid hotkey on save. Nothing out of scope was bundled.

### Correctness
I traced the paths that worried me and they hold up:

- `ClassifyConfiguredHotkey` validates with `HotkeyValidator.Validate` and `TryRegisterCore` then calls `Hotkey.Parse` unguarded — but every string `Validate` accepts also parses without throwing (`Validate` delegates to the same `Hotkey.Parse`, and `Parse` treats a comma as a token separator), so no exception can slip through.
- At startup the failure dialog is shown after `_window.Activate()`, so the `XamlRoot` guard in `ShowHotkeyBindingFailuresAsync` does not silently reintroduce the bug; the failure beep also plays before the dialog guard as a non-visual cue.
- The redundant second `RegisterRouterHotkeys()` call in `App.OnLaunched` was correctly removed — `AttachHotkeyManager` performs the one registration and returns its failures.
- Accessibility: the dialog sets `AutomationProperties.Name` and `HelpText`, and each failure names the action or prompt, not just the key string.

### Tests
The 15 new tests in `Mutation.Tests/HotkeyBindingFailureTests.cs` cover valid/invalid/empty classification, message formatting (with and without hotkey text, multiple failures), and router-result mapping including the missing-error-message fallback. Registration conflicts need a live window handle and are reasonably left to manual testing, as the PR notes.

### Regressions
None found. All callers of the changed methods (`App.OnLaunched`, `ApplyLiveSettings`, `PromptLibraryController`) were updated consistently, and `ClearAllForRebind` still unregisters everything before a re-registration pass.

### Minimality
All changes serve the story. One piece of leftover state was introduced: after the old startup dialog was removed, `HotkeyManager.FailedRegistrations` and `CoreFailedRegistrations` had no readers left and were maintained as write-only lists.

### Fixes applied
- 0ea904d — removed the now-dead `FailedRegistrations` / `CoreFailedRegistrations` state, the router/core failure bookkeeping lists, and the unused `isRouter` parameter of `RegisterHotkeyInternal`. Full suite re-run: 538/538 green.

No issues remaining.

---
Reviewed at 0ea904d20defc2f9f5b4d51d892fc5a9089598c9
🤖 Reviewed with Claude Code

PR #192, opened 2026-07-04
Surface settings save failures with InfoBar, UIA announcement, and beep

jfbosch:
## Summary

When saving settings failed (a locked file, read-only disk, or serialization error), the Settings dialog silently refused to close: the exception was swallowed, and the only trace was a debug-output line. A screen-reader user pressing Save had no way to tell the press even registered.

This change surfaces the failure three ways, keeping the dialog open so the user can retry or copy their values:

- An error InfoBar inside the dialog shows what went wrong, with guidance to fix and press Save again or Cancel to discard.
- A UIA notification (assertive, interrupting) announces the failure to screen readers, using the existing `StatusAnnouncement` mapping.
- The failure beep plays, matching the app's audio-feedback conventions.

The same feedback now applies when the "Open Mutation.json" button fails to launch an external editor.

Message composition lives in a new WinUI-free helper, `SettingsFailureFeedback`, which unwraps to the innermost exception message (the actionable cause) and falls back to generic text when no message is available.

Closes #165

## Test plan

- 5 new unit tests in `SettingsFailureFeedbackTests` cover message composition: retry guidance included, innermost exception unwrapping, null-exception and blank-message fallbacks, and open-JSON detail-only text.
- Full quality gate: `dotnet test --configuration Release` — 605 passed, 0 failed.
- Manual check (recommended for a reviewer with the app running): make `Mutation.json` read-only, open Settings, change a value, press Save — the dialog should stay open, show the error InfoBar, announce it via the screen reader, and play the failure beep.

### DIFF
--- a/Mutation.Tests/SettingsFailureFeedbackTests.cs
+++ b/Mutation.Tests/SettingsFailureFeedbackTests.cs
@@ -1,0 +1,59 @@
+using System;
+using System.IO;
+using Mutation.Ui.Views.SettingsUi;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class SettingsFailureFeedbackTests
+{
+	[Fact]
+	public void ComposeMessage_IncludesExceptionMessageAndRetryGuidance()
+	{
+		string message = SettingsFailureFeedback.ComposeMessage(
+			new IOException("The file is locked by another process."));
+
+		Assert.StartsWith("The file is locked by another process.", message);
+		Assert.Contains("Your changes were not saved.", message);
+		Assert.Contains("press Save again", message);
+	}
+
+	[Fact]
+	public void ComposeMessage_UnwrapsToInnermostException()
+	{
+		var wrapped = new InvalidOperationException(
+			"Serialization failed.",
+			new UnauthorizedAccessException("Access to the path is denied."));
+
+		string message = SettingsFailureFeedback.ComposeMessage(wrapped);
+
+		Assert.StartsWith("Access to the path is denied.", message);
+		Assert.DoesNotContain("Serialization failed.", message);
+	}
+
+	[Fact]
+	public void ComposeMessage_NullException_UsesFallbackText()
+	{
+		string message = SettingsFailureFeedback.ComposeMessage(null);
+
+		Assert.StartsWith("An unknown error occurred.", message);
+		Assert.Contains("Your changes were not saved.", message);
+	}
+
+	[Fact]
+	public void ComposeMessage_BlankExceptionMessage_UsesFallbackText()
+	{
+		string message = SettingsFailureFeedback.ComposeMessage(new Exception("   "));
+
+		Assert.StartsWith("An unknown error occurred.", message);
+	}
+
+	[Fact]
+	public void ComposeOpenJsonMessage_ReturnsInnermostDetailOnly()
+	{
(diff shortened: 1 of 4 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR fixes the silent Settings save failure (#165): when `SaveSettingsToFile`, `CommitInto`, `_onLiveApply`, or the "Open Mutation.json" launch throws, the dialog now surfaces the failure via an error `InfoBar`, an assertive UIA notification, and the failure beep, while keeping the dialog open. It does this correctly.

### Non-compliance
None. Build (`dotnet build --configuration Release`) and the full suite (`dotnet test --configuration Release`, 605 passed / 0 failed) are green. Accessibility is preserved and improved.

### Story alignment
Fully matches the issue. Both failure sites named in #165 — `SettingsDialog_PrimaryButtonClick` (save) and `OpenJsonButton_Click` (open JSON) — are covered. Failures now show an in-dialog `InfoBar`, raise a UIA notification, and play the failure beep, with the dialog staying open so the user can retry or Cancel to discard — exactly the suggested fix.

### Correctness
The logic is sound. `SettingsFailureFeedback.ExtractDetail` walks to the innermost `InnerException` — the actionable cause — and falls back to "An unknown error occurred." on null/blank, verified by the substituted arithmetic in the tests. `ShowFailure` (`SettingsDialog.xaml.cs:233`) mirrors the established `MainWindow.AnnounceStatus` pattern exactly: `CreatePeerForElement` (so the notification fires even before a peer exists) plus `RaiseNotificationEvent` with `ImportantMostRecent` for `Error` severity, which interrupts the screen reader — appropriate for a save failure. `ComposeText(title, message)` is never empty here since both titles are non-empty constants.

### Accessibility
Correctly handled for the screen-reader user: `AutomationProperties.SetName`/`SetHelpText` on the `InfoBar` plus the assertive, interrupting UIA notification. The `Severity="Error"` `InfoBar` gives sighted/ZoomText users the visible cue. No keyboard trap or focus loss introduced — the bar is inserted as its own grid row (`Grid.Row="2"`), with the button panel cleanly shifted to `Grid.Row="3"`.

### Tests
Sufficient for the extracted logic. `SettingsFailureFeedbackTests` covers the five meaningful branches: retry-guidance inclusion, innermost-exception unwrapping, null-exception fallback, blank-message fallback, and open-JSON detail-only text. The `ShowFailure` UI method is not unit-tested, which is consistent with the codebase convention of extracting testable text composition into a WinUI-free helper (as `StatusAnnouncement` does) and leaving the thin WinUI glue untested.

### Regressions
None. The new `SettingsFailureFeedback` type is additive; `ShowFailure` is a new private method; the two catch blocks gain a feedback call without altering existing control flow (`args.Cancel = true` is retained on save failure). The XAML row-index shift is self-contained.

### Minimality
Every changed line serves the fix. New helper (30 lines), its tests (59 lines), one InfoBar row in XAML, and the `ShowFailure` method plus two call sites. No unrelated refactors or formatting churn. The file follows the sibling `Mutation.Ui.Views.SettingsUi` namespace convention.

### Fixes applied

No issues remaining.

---
Reviewed at 043fe4e2d200dbaaea878445690e1a02586c2b7b
🤖 Reviewed with Claude Code

PR #193, opened 2026-07-04
Make settings search screen-reader accessible; name search box and Advanced toggle

jfbosch:
## Summary

The settings dialog has a search box that filters a nine-category settings panel. Until now it signalled matches only by dimming non-matching sections to 35% opacity. Dimmed controls keep their tab stops and read identically to a screen reader, so search results were invisible to the app's primary user (a blind developer), and 35% opacity is poor contrast under ZoomText. Two pieces of dialog chrome were also unlabeled.

This change makes the search perceivable without sight:

- **Non-matching sections are now collapsed, not dimmed.** Collapsing removes them from the tab order and the screen-reader tree, so what remains after a search is exactly the set of matches — reachable by keyboard and announced by the screen reader. An empty query restores every section.
- **A status line under the search box announces the result.** It is a visible caption that doubles as a polite live region, so each search updates a message like "3 of 9 categories match. 2 matching sections shown on this page." without moving focus. A visible status (rather than a hidden one) matches the project preference for exposed controls.
- **The search box and the Advanced toggle now have accessible names** (`AutomationProperties.Name`), so they no longer read as unnamed controls.

The match-counting text is composed by a new pure helper (`SettingsSearchStatus`) with no UI dependencies, so it is unit-tested in isolation — the same pattern the codebase already uses for `StatusAnnouncement`.

## Test plan

- `dotnet test --configuration Release` — all 618 tests pass, including six new `SettingsSearchStatusTests` covering the empty query, no-match, categories-only, both-counts, singular/plural section wording, and zero-section cases.
- Manual (screen reader): open Settings, type a query, and confirm non-matching categories and sections disappear from tab navigation and the status line is announced; clear the query and confirm all sections return; confirm the search box and Advanced toggle announce their names.

Closes #166

### DIFF
--- a/Mutation.Tests/SettingsSearchStatusTests.cs
+++ b/Mutation.Tests/SettingsSearchStatusTests.cs
@@ -1,0 +1,48 @@
+using Mutation.Ui.Views.SettingsUi;
+
+namespace Mutation.Tests;
+
+public class SettingsSearchStatusTests
+{
+	[Theory]
+	[InlineData(null)]
+	[InlineData("")]
+	[InlineData("   ")]
+	public void Compose_EmptyQuery_ReturnsEmpty(string? query)
+	{
+		Assert.Equal(string.Empty, SettingsSearchStatus.Compose(query, 9, 9, 5));
+	}
+
+	[Fact]
+	public void Compose_NoCategoryMatches_SaysNoMatch()
+	{
+		Assert.Equal("No categories match.", SettingsSearchStatus.Compose("zzz", 0, 9, null));
+	}
+
+	[Fact]
+	public void Compose_NoPageLoaded_ReportsCategoriesOnly()
+	{
+		Assert.Equal("3 of 9 categories match.", SettingsSearchStatus.Compose("audio", 3, 9, null));
+	}
+
+	[Fact]
+	public void Compose_WithPageMatches_ReportsBothCounts()
+	{
+		Assert.Equal("3 of 9 categories match. 2 matching sections shown on this page.",
+			SettingsSearchStatus.Compose("audio", 3, 9, 2));
+	}
+
+	[Fact]
+	public void Compose_SingleSectionMatch_UsesSingular()
+	{
+		Assert.Equal("1 of 9 categories match. 1 matching section shown on this page.",
+			SettingsSearchStatus.Compose("mute", 1, 9, 1));
+	}
+
+	[Fact]
+	public void Compose_ZeroSectionsOnMatchingPage_ReportsZero()
+	{
+		Assert.Equal("2 of 9 categories match. 0 matching sections shown on this page.",
+			SettingsSearchStatus.Compose("key", 2, 9, 0));
+	}
+}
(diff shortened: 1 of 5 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR makes the settings search perceivable to a screen-reader/ZoomText user by collapsing non-matching sections (removing them from the tab order and UIA tree) instead of dimming them, adding a polite live-region status line under the search box, and giving the search box and Advanced toggle accessible names. It implements the linked issue correctly and the change is clean, minimal, and well-tested.

### Non-compliance
None. `dotnet build --configuration Release` succeeds with 0 warnings/0 errors; `dotnet test --configuration Release` is green (618/618). No secrets. The change strengthens accessibility rather than regressing it, and the status line is exposed as a visible caption (a configurable/visible control) rather than a hidden default.

### Story alignment
Fully aligned with issue #166. All three asks are delivered: `AutomationProperties.Name` on the `SearchBox` ("Search settings") and the `AdvancedToggle` ("Advanced settings"); non-matching sections are now `Visibility.Collapsed` instead of `Opacity = 0.35`; and the match count is announced through a `LiveSetting="Polite"` live region. Nothing out of scope was added.

### Correctness
Traced the search flow end to end. `SettingsSearchHelper.ApplyFilter` now returns the match count and toggles `Visibility` (`SettingsSearchHelper.cs:17-42`); both call sites — the deferred re-apply in `LoadSelectedPage` (`SettingsDialog.xaml.cs:181`) and `FilterActivePage` (`SettingsDialog.xaml.cs:211`) — consume the new return correctly, and the removed `HighlightActivePage` has no lingering references. The zero-category path is handled: clearing the category list drives `CategoryList.SelectedIndex` to -1 → `ShowEmptyState`, `FilterActivePage` returns null, and `SettingsSearchStatus.Compose` yields "No categories match." The `Compose` helper is a pure function with correct empty/no-match/categories-only/both-count branches and singular/plural section wording.

Minor observation (non-blocking, not filed): the status `TextBlock` starts `Collapsed` and is flipped to `Visible` in the same update that sets its text. Subsequent updates announce reliably; the very first Collapsed→Visible transition *may* not raise a LiveRegionChanged event on some WinUI builds. The section collapse itself is still perceivable via tab-order change, so this is at most a small first-announcement nuance, not a defect — flagging only for awareness.

### Tests
`SettingsSearchStatusTests` adds six cases covering the empty query, no-match, categories-only, both-counts, singular-section, and zero-section paths — good coverage of the pure helper, matching the existing `StatusAnnouncement` test pattern. The visual collapse logic in `ApplyFilter` is not unit-tested, which is reasonable: it depends on a live `VisualTreeHelper` tree that cannot be exercised in a headless test.

### Regressions
None found. The signature change to `ApplyFilter` is fully propagated; no other callers exist. Unrelated dialog behavior (sizing, save, Escape-to-clear) is untouched.

### Minimality
Every changed line serves the stated purpose. The XAML wrapping of the search box in a `StackPanel` is required to host the new status caption. No unrelated refactors or formatting churn.

### Fixes applied

No issues remaining.

---
Reviewed at b65747a3f8b99d275c0be9fb9af8570a632f5926
🤖 Reviewed with Claude Code

PR #194, opened 2026-07-04
Queue dialogs requested while another is open instead of dropping them (#167)

jfbosch:
Closes #167

## What this fixes

`ShowDialogAsync` refused to show a dialog while another one was already
open: it returned immediately and the new dialog was never queued or shown.
Because every error dialog (`ShowErrorDialog`) and the audio-session error
handler funnel through this method, a background transcription or OCR
failure that fired while the Settings dialog — or any content dialog — was
open was discarded entirely. For the error paths that call `ShowErrorDialog`
directly (screenshot, OCR, speech-to-text, LLM), nothing was shown and
nothing was announced, so a blind user was left with the operation appearing
to have produced nothing.

## What changed

- **New `DialogQueue<TResult>`** (a small, dependency-free helper): a serial
  first-in-first-out queue that shows modal dialogs one at a time. A dialog
  requested while another is open is queued and shown when the current one
  closes — never dropped. Callers still `await` and receive the dialog's
  real result, whether it ran immediately or waited its turn. A show
  operation that throws faults only its own result; the queue keeps draining
  the rest.
- **`ShowDialogAsync`** now enqueues through this queue instead of dropping
  the dialog on a busy flag.
- **Immediate feedback on deferral.** When a dialog has to wait behind
  another, an accessibility notification is raised right away with its title
  and message, and error dialogs also play the failure beep. The user learns
  a dialog is pending immediately rather than sitting in silence until the
  blocking dialog closes.
- The delete-prompt confirmation previously relied on the old
  "another dialog open → treated as cancel" behavior; its comment is updated.
  A confirmation requested while busy is now queued and actually shown, and
  deletion still requires an explicit Delete click, so nothing is deleted
  without confirmation.

## How the issue's suggestion is addressed

The issue suggested either queueing pending dialogs or, at minimum, routing
the suppressed dialog's title and message through an accessibility
notification plus the failure beep. This does both: dialogs are queued (the
primary fix) **and** deferral is announced immediately with the failure beep
for errors.

## Test plan

- New unit tests in `DialogQueueTests` cover: a single dialog showing
  immediately and returning its result; a second dialog queued behind the
  first and shown in FIFO order only after the first closes; `IsBusy` /
  `PendingCount` transitions; result propagation; and a faulting show
  operation faulting only its own task while the queue keeps draining.
- `dotnet test --configuration Release` — all 614 tests pass.
- The dialog-showing itself (WinUI `ContentDialog.ShowAsync`) is UI-thread
  bound and not unit-testable in isolation; the queueing logic it depends on
  is fully covered.

### DIFF
--- a/Mutation.Tests/DialogQueueTests.cs
+++ b/Mutation.Tests/DialogQueueTests.cs
@@ -1,0 +1,88 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Mutation.Ui.Core;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class DialogQueueTests
+{
+	[Fact]
+	public async Task SingleDialog_ShowsImmediately_AndReturnsItsResult()
+	{
+		var queue = new DialogQueue<int>();
+
+		int result = await queue.EnqueueAsync(() => Task.FromResult(42));
+
+		Assert.Equal(42, result);
+		Assert.False(queue.IsBusy);
+		Assert.Equal(0, queue.PendingCount);
+	}
+
+	[Fact]
+	public async Task SecondDialog_IsQueuedAndShownWhenFirstCloses_InFifoOrder()
+	{
+		var queue = new DialogQueue<int>();
+		// Default TaskCompletionSource (inline continuations) so the queue's
+		// pump advances synchronously on this thread the moment a gate is
+		// completed — keeps the test deterministic and single-threaded, which
+		// mirrors the queue's UI-dispatcher affinity.
+		var gateA = new TaskCompletionSource<int>();
+		var gateB = new TaskCompletionSource<int>();
+		var shownOrder = new List<string>();
+
+		Task<int> first = queue.EnqueueAsync(() => { shownOrder.Add("A"); return gateA.Task; });
+		Task<int> second = queue.EnqueueAsync(() => { shownOrder.Add("B"); return gateB.Task; });
+
+		// A is on screen; B has not started and is waiting behind it.
+		Assert.Equal(new[] { "A" }, shownOrder);
+		Assert.True(queue.IsBusy);
+		Assert.Equal(1, queue.PendingCount);
+
+		// Closing A lets the queue show B next.
+		gateA.SetResult(1);
+		Assert.Equal(new[] { "A", "B" }, shownOrder);
+		Assert.Equal(0, queue.PendingCount);
+		Assert.True(queue.IsBusy);
+
+		// Closing B drains the queue.
+		gateB.SetResult(2);
+		Assert.False(queue.IsBusy);
+
+		Assert.Equal(1, await first);
+		Assert.Equal(2, await second);
+	}
+
+	[Fact]
+	public async Task FaultingShow_FaultsOnlyItsOwnTask_AndTheQueueKeepsDraining()
+	{
+		var queue = new DialogQueue<int>();
+		var boom = new InvalidOperationException("boom");
+
+		Task<int> failing = queue.EnqueueAsync(() => Task.FromException<int>(boom));
+		Task<int> following = queue.EnqueueAsync(() => Task.FromResult(7));
+
+		var thrown = await Assert.ThrowsAsync<InvalidOperationException>(() => failing);
+		Assert.Same(boom, thrown);
+		Assert.Equal(7, await following);
+		Assert.False(queue.IsBusy);
+		Assert.Equal(0, queue.PendingCount);
+	}
+
+	[Fact]
+	public void EnqueueAsync_NullShowOperation_Throws()
+	{
+		var queue = new DialogQueue<int>();
+
+		// The null guard throws synchronously before any task is created, so
+		// Assert.Throws is correct despite EnqueueAsync's Task return type; the
+		// analyzer cannot see that the throw is synchronous, so suppress it.
+#pragma warning disable xUnit2014
+		Assert.Throws<ArgumentNullException>(() =>
+		{
+			queue.EnqueueAsync(null!);
+		});
+#pragma warning restore xUnit2014
+	}
+}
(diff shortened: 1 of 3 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR replaces the `_isDialogOpen` "drop the dialog if one is already open" guard with a serial FIFO `DialogQueue<TResult>`, so a dialog requested while another is open is now queued and shown when the current one closes rather than silently discarded. It correctly implements issue #167 and does so accurately.

### Non-compliance
None. Build is clean (`dotnet build --configuration Release`, 0 warnings) and the full suite is green (`dotnet test --configuration Release`, 614 passed). No accessibility regression — the change adds a UIA notification plus failure beep and preserves `AutomationProperties` on every dialog. No hidden defaults; the behavior is the fix itself, not a hidden toggle.

### Story alignment
Fully addresses #167. The issue asked to either queue pending dialogs **or**, at minimum, route the suppressed dialog's title/message through a UIA notification plus the failure beep. This does both: queueing is the primary fix (`ShowDialogAsync` now enqueues via `_dialogQueue`), and deferral is announced immediately via `AnnouncePendingDialog` with the failure beep for error dialogs (`ShowErrorDialog`).

### Correctness
`DialogQueue.PumpAsync` is sound under the documented single-UI-thread contract: `EnqueueAsync` starts the pump only when `!_isPumping`, subsequent enqueues ride the running pump, and the `_isPumping = false` reset sits in a no-`await` window after the drain so no interleaving can strand the flag (`Mutation.Ui/Core/DialogQueue.cs:57-73`). A faulting show operation faults only its own task and the pump keeps draining (`:64-69`). `RunContinuationsAsynchronously` correctly keeps the caller's continuation off the pump's stack. Onboarding sequencing (`MainWindow.Onboarding.cs:23-28`, `:131-132`) still works because each `ShowDialogAsync` await completes only when its dialog closes, and the existing `YieldToDispatcherAsync` still separates back-to-back dialogs.

Note (pre-existing, out of scope): direct `ContentDialog.ShowAsync()` calls in `App.xaml.cs` and `PromptEditorWindow.xaml.cs` bypass this queue, exactly as they bypassed the old `_isDialogOpen` flag. Those are separate-window / terminal (startup/shutdown/unhandled-exception) dialogs, so this is neither a regression nor in scope for #167.

### Tests
`DialogQueueTests` covers the meaningful logic: immediate show + result propagation, FIFO ordering with a second dialog held behind the first, `IsBusy`/`PendingCount` transitions, fault isolation, and the synchronous null guard. The WinUI `ContentDialog.ShowAsync` call itself is UI-thread-bound and not unit-testable in isolation, which the PR correctly acknowledges; the queueing logic it depends on is fully exercised.

### Regressions
None found. The delete-prompt confirmation still requires an explicit Delete click — a confirmation requested while busy is now queued and actually shown instead of being treated as an implicit cancel, and its comment is updated to match (`MainWindow.xaml.cs:2357-2367`).

### Minimality
Every change is necessary for the fix. No unrelated refactors or formatting churn; tabs are used per repo convention.

### Fixes applied

No issues remaining.

---
Reviewed at 415af92b0d442bb98e50713f4b4fc794d9d0023d
🤖 Reviewed with Claude Code

jfbosch:
## Review by Claude

**Verdict: Approved**

The PR replaces `ShowDialogAsync`'s drop-on-busy guard with a serial FIFO `DialogQueue<TResult>`, so a dialog requested while another is open is queued and shown when the current one closes instead of vanishing. It correctly resolves issue #167 and does what it claims.

### Non-compliance
None. Solution builds and `dotnet test --configuration Release` is green (614 passed). No accessibility regression — deferral now raises an explicit UIA notification (title + message) and plays the failure beep for errors, which is strictly more accessible than the old silent drop. No hidden defaults, no secrets.

### Story alignment
Fully addresses the issue. The suggested fix offered two options (queue, or at minimum announce title/message + beep); this does both — queueing is the primary fix and deferral is announced immediately with the failure beep for error dialogs.

### Correctness
`DialogQueue` is sound. It is single-thread-affine to the UI dispatcher (documented), which matches the pre-existing `_isDialogOpen` assumption, so it introduces no new threading concern. The pump drains FIFO, one dialog at a time; a faulting show operation faults only its own `TaskCompletionSource` and the queue keeps draining (`PumpAsync`, `Core/DialogQueue.cs:57`). `RunContinuationsAsynchronously` on the caller's completion avoids resuming an awaiting UI handler nested inside the pump loop. The `_isDialogOpen` field is fully removed with no dangling references.

One latent observation, not blocking and not a regression: because dialogs are now serialized, a dialog awaited from *inside* another open dialog's handler (a true nested-modal flow) would wait for the outer to close — and if the outer's close depended on the inner, it would hang rather than, as before, drop the inner and return `None`. No current code path nests dialogs this way (onboarding, the OCR-warning->Settings flow, delete-prompt confirmation, and all `ShowErrorDialog` sites are top-level and sequential), so there is nothing to fix today; worth keeping in mind if a nested flow is ever added.

### Tests
`DialogQueueTests` covers the meaningful behaviours: immediate show + result propagation, a second dialog queued behind the first shown in FIFO order only after the first closes, `IsBusy`/`PendingCount` transitions, fault isolation, and the null-argument guard. The gated tests use inline-continuation `TaskCompletionSource` to keep the pump deterministic and single-threaded, mirroring the dispatcher affinity. The WinUI `ContentDialog.ShowAsync` call is UI-thread-bound and not unit-testable in isolation; the queueing logic it depends on is fully covered.

### Regressions
None found. The delete-prompt confirmation's stale "another dialog open -> treated as cancel" comment was updated to match the new queued behaviour, and deletion still requires an explicit Delete click. All other `ShowDialogAsync`/`ShowErrorDialog` call sites are unaffected.

### Minimality
Every change is in service of the fix: the new queue, its wiring into `ShowDialogAsync`, the deferral announcement/beep, and the corrected comment. No unrelated refactors.

### Fixes applied

No issues remaining.

---
Reviewed at 415af92b0d442bb98e50713f4b4fc794d9d0023d
🤖 Reviewed with Claude Code

PR #195, opened 2026-07-04
Guard recorder start and roll back a failed start (#178)

jfbosch:
## What this fixes

Closes #178

Starting a dictation recording could leave the app permanently stuck in the "recording" state. The manager that owns the recorder treats "a recorder object exists" as the definition of "we are recording", but it created that recorder object and assigned it to its field *before* actually starting the audio capture. If the capture failed to start — for example the microphone was unplugged, the audio backend threw an error, or the output file could not be opened — the manager was left holding a recorder object that was doing nothing, yet it kept reporting that a recording was in progress. The Stop button stayed active, and pressing it tried to transcribe a zero-byte file, producing a confusing error.

The same code also had no guard against being asked to start a second recording while one was already running. A second start would silently replace the live recorder without stopping it, leaving the microphone open and the in-progress audio file truncated and unrecoverable.

## What changed

- `SpeechToTextManager.StartRecordingAsync` now starts the recorder into a local variable and only commits it to the field once the start succeeds. If the start throws, it disposes the local recorder, clears the in-progress session, and deletes the empty session file it had just created — so the app returns to a clean idle state instead of a phantom "recording" state.
- Starting a recording while one is already live is now rejected with a clear error, before any session file is created, so a live capture is never overwritten or orphaned.
- To make these paths testable without real microphone hardware, the recorder contract the manager depends on was extracted into a new `IAudioRecorder` interface (implemented by the existing `AudioRecorder`), and the manager now obtains its recorder through an injectable factory. Production behaviour is unchanged; the default factory builds the same concrete recorder as before.

## Test plan

New unit tests in `Mutation.Tests/SpeechToTextManagerStartGuardTests.cs` cover the three cases, using a fake recorder injected through the new factory seam:

- A recorder whose start fails leaves the manager reporting *not* recording, with no current session and no leftover empty session file on disk.
- After a failed start, a subsequent start succeeds normally (the manager is not wedged).
- Starting while a recording is already live throws, and the live recorder is neither replaced nor disposed.

Full gate: `dotnet test --configuration Release` — 626 passed, 0 failed.

### DIFF
--- a/CognitiveSupport/AudioRecorder.cs
+++ b/CognitiveSupport/AudioRecorder.cs
@@ -6,7 +6,7 @@ using System.IO;
 
 namespace CognitiveSupport;
 
-public class AudioRecorder : IDisposable
+public class AudioRecorder : IAudioRecorder
 {
 	private WaveInEvent? _waveIn;
 	private OpusEncoder? _encoder;
--- a/CognitiveSupport/IAudioRecorder.cs
+++ b/CognitiveSupport/IAudioRecorder.cs
@@ -1,0 +1,29 @@
+using System;
+
+namespace CognitiveSupport;
+
+/// <summary>
+/// The recording contract <c>SpeechToTextManager</c> depends on. Extracting it
+/// from the concrete <see cref="AudioRecorder"/> (which drives real NAudio
+/// hardware) lets the manager be unit-tested in isolation with a recorder that
+/// can fail a start on demand.
+/// </summary>
+public interface IAudioRecorder : IDisposable
+{
+	/// <summary>
+	/// After <see cref="StopRecording"/>, the trimmed speech duration in seconds
+	/// when silence stripping was active; null when stripping was disabled.
+	/// </summary>
+	double? TrimmedSpeechSeconds { get; }
+
+	/// <summary>
+	/// First exception (if any) that occurred while encoding audio on the capture
+	/// thread, captured there rather than thrown so it does not terminate the
+	/// process. Callers surface it after <see cref="StopRecording"/>.
+	/// </summary>
+	Exception? CaptureException { get; }
+
+	void StartRecording(int captureDeviceIndex, string outputFile, SilenceTrimmerOptions? silenceOptions = null);
+
+	void StopRecording();
+}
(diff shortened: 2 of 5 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR fixes issue #178: `SpeechToTextManager.StartRecordingAsync` previously assigned the recorder to its field before the capture started, so a failed start left the app permanently reporting "recording" with no active capture, and a second start could silently orphan a live recorder. The fix starts the recorder into a local, commits the field only on success, rolls back cleanly on failure, and guards against a double start. It does this correctly.

### Non-compliance
None. `dotnet build --configuration Release` and `dotnet test --configuration Release` both pass (626 passed, 0 failed). No UI surface changed, so there is no accessibility impact. No secrets committed.

### Story alignment
Implements exactly what #178 describes and nothing more: the already-recording guard runs before the session file is created (`SpeechToTextManager.cs:166`), the recorder is started into a local and committed to the field only on success (`SpeechToTextManager.cs:184-200`), and a failed start disposes the local recorder, clears `_currentRecordingSession`, and deletes the just-created empty session file (`SpeechToTextManager.cs:189-198`). The `IAudioRecorder` extraction and factory seam are the minimal change needed to make these paths testable without real hardware.

### Correctness
Traced the failure path: on a `StartRecording` throw, the local recorder is disposed, the session is nulled under `_sessionLock`, the file is removed via `TryDeleteSessionFile`, and the exception rethrows — leaving `_audioRecorder` null so `Recording` reports false. The guard and field access are all serialized by `AudioRecorderLock` held across the method, so there is no race on `_audioRecorder`. `TryDeleteSessionFile` swallows only the expected IO exceptions, matching the pattern already used in `CreateSessionFileAsync`.

### Tests
`SpeechToTextManagerStartGuardTests` covers the three meaningful cases with a clean injected fake: failed start leaves the manager idle with no orphan session file, a subsequent start recovers, and a start while already recording throws without touching the live recorder. Boundary/error paths are exercised directly.

### Regressions
`SpeechToTextState` was updated to `Func<IAudioRecorder?>` and `AudioRecorder` now implements `IAudioRecorder`; the only consumers are these files. Production wiring is unchanged — the public constructor still builds a concrete `AudioRecorder` through the default factory.

### Minimality
Every change is necessary for the fix or its testability. No unrelated refactors or formatting churn.

### Fixes applied

No issues remaining.

---
Reviewed at 7ea859a556e3a2d423ca4ca5771c7bb881b0088f
🤖 Reviewed with Claude Code

PR #196, opened 2026-07-04
Run the mic mute toggle off the UI thread (#171)

jfbosch:
## What this changes

The microphone mute-toggle hotkey used to do all of its work directly on the
UI thread: it writes the new mute state to every capture device over the
Windows audio COM API, reads each one back to confirm it took effect, and — if
any write fails — re-enumerates every audio device and retries. All of that ran
on the thread that also draws the window and drives the screen reader, and it
held the same lock the operating system uses to notify the app when a
microphone is plugged in or removed. So pressing the mute hotkey while a device
was being hot-plugged could freeze the UI for tens to hundreds of milliseconds
— worst exactly on the failure/retry path the verify-and-retry logic exists
for.

This moves that work off the UI thread. A new `MicrophoneMuteToggleCoordinator`
runs the toggle (writes, read-back verification, and any device
re-enumeration) on a background thread and hands only the confirmed result back
to the UI, where the icon, status message, and beep are updated as before. The
coordinator also coalesces overlapping presses: while one toggle is in flight,
a second press is dropped rather than queued, so a fast double-press can't stack
two flips and leave the mic where it started.

## Scope

The change targets the **mute toggle** specifically. The microphone-level
slider and the record-start path share the same "COM write on the UI thread"
pattern, but moving those off-thread safely needs a different, ordering-aware
design (each slider tick must not overtake a later one). That is filed as a
separate follow-up rather than bundled here.

## Behaviour notes

- The confirmed-state semantics are unchanged: the mute icon and status still
  reflect only a state that was written and read back successfully, so a failed
  mute never shows the muted icon.
- On the normal (healthy-device) path the extra background hop is imperceptible;
  the win is that the failure/retry path no longer blocks the UI or the screen
  reader.
- Accessibility: the status InfoBar announcement and the mute/unmute/failure
  beeps still fire, now from the resumed UI thread after the confirmed result
  comes back.

## Test plan

- `dotnet test --configuration Release` — all 628 tests pass (5 new).
- New `MicrophoneMuteToggleCoordinatorTests` cover: the toggle runs off the
  calling thread, its result is returned, overlapping requests are coalesced to
  `null`, a fresh toggle is allowed after the previous one completes, and a null
  toggle delegate is rejected.
- Existing `MuteStateControllerTests` continue to cover the write / verify /
  retry logic the coordinator drives.

Closes #171

### DIFF
--- a/Mutation.Tests/MicrophoneMuteToggleCoordinatorTests.cs
+++ b/Mutation.Tests/MicrophoneMuteToggleCoordinatorTests.cs
@@ -1,0 +1,102 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Mutation.Ui.Core;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class MicrophoneMuteToggleCoordinatorTests
+{
+	[Fact]
+	public void Constructor_NullToggle_Throws()
+	{
+		Assert.Throws<ArgumentNullException>(() => new MicrophoneMuteToggleCoordinator(null!));
+	}
+
+	[Fact]
+	public async Task ToggleAsync_ReturnsToggleResult()
+	{
+		var expected = new MuteToggleResult(Success: true, IsMuted: true);
+		var coordinator = new MicrophoneMuteToggleCoordinator(() => expected);
+
+		var result = await coordinator.ToggleAsync();
+
+		Assert.Equal(expected, result);
+		Assert.False(coordinator.IsToggling);
+	}
+
+	[Fact]
+	public async Task ToggleAsync_RunsToggleOffTheCallingThread()
+	{
+		// The delegate signals when it starts and then blocks. If the toggle ran
+		// synchronously on this thread, ToggleAsync would block inside the delegate
+		// and never hand back a task — so simply reaching the asserts below, with
+		// "started" already signalled, proves the work runs on another thread.
+		var started = new ManualResetEventSlim(false);
+		var release = new ManualResetEventSlim(false);
+		var coordinator = new MicrophoneMuteToggleCoordinator(() =>
+		{
+			started.Set();
+			release.Wait();
+			return new MuteToggleResult(Success: true, IsMuted: true);
+		});
+
+		var toggle = coordinator.ToggleAsync();
+
+		Assert.True(started.Wait(TimeSpan.FromSeconds(5)), "toggle did not start on a background thread");
+		Assert.False(toggle.IsCompleted);
+		Assert.True(coordinator.IsToggling);
+
+		release.Set();
+		var result = await toggle;
+
+		Assert.Equal(new MuteToggleResult(Success: true, IsMuted: true), result);
+		Assert.False(coordinator.IsToggling);
+	}
+
+	[Fact]
+	public async Task ToggleAsync_CoalescesOverlappingRequests()
+	{
+		var release = new ManualResetEventSlim(false);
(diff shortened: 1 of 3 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR moves the microphone mute-toggle's COM writes, read-back verification, and device re-enumeration off the UI thread via a new `MicrophoneMuteToggleCoordinator`, marshalling only the confirmed result back to the UI. It does this correctly and is well-scoped.

### Non-compliance
None. `dotnet build --configuration Release` succeeds with 0 warnings and `dotnet test --configuration Release` is green (628 passed, 5 new). No accessibility regression — the status InfoBar announcement and mute/unmute/failure beeps still fire on the resumed UI thread. No hidden defaults, no secrets.

### Story alignment
Matches issue #171 exactly: the toggle/verify/retry pipeline now runs on a background thread (`Task.Run`) and only the confirmed `MuteToggleResult` returns to the UI. The mic-level slider and record-start paths — which the issue notes share the pattern but need an ordering-aware design — are correctly deferred to a follow-up rather than bundled here.

### Correctness
- Coalescing is correct: `Interlocked.CompareExchange(ref _inFlight, 1, 0)` in `MicrophoneMuteToggleCoordinator.ToggleAsync` runs synchronously before the first `await`, so an overlapping second press is dropped (returns `null`) rather than queued, and can't stack two flips.
- The continuation resumes on the captured context. `BtnToggleMic_Click` is enqueued on the `DispatcherQueue` on the hotkey path and fired on the UI thread on the button path, so the WinUI `DispatcherQueueSynchronizationContext` is present and `UpdateMicrophoneToggleVisuals`/`ShowStatus`/`BeepPlayer.Play` all run on the UI thread (`MainWindow.xaml.cs:566-582`).
- No `async void` crash risk: `MuteStateController.Toggle()` catches per-endpoint COM exceptions internally (`MuteStateController.cs:78-88`) and returns a result instead of throwing, and the coordinator's `finally` resets the in-flight flag on every path.
- Nullable handling is clean: the `is not MuteToggleResult result` pattern-match correctly treats a coalesced (`null`) toggle as a no-op, leaving the UI showing the last confirmed state.

Observation (non-blocking): this is the first path to call the NAudio CoreAudio COM objects from a thread-pool thread rather than the UI thread. The issue explicitly sanctions running the pipeline off-thread, and these interfaces marshal across apartments, so this is the intended design — flagging only so it's on the record when the slider/record-start follow-up moves the same COM calls off-thread.

### Tests
`MicrophoneMuteToggleCoordinatorTests` covers the meaningful behaviors: work runs off the calling thread, the result is returned, overlapping requests coalesce to `null`, a fresh toggle is allowed after the previous completes, and a null delegate is rejected. The underlying write/verify/retry logic remains covered by `MuteStateControllerTests`.

### Regressions
`BtnToggleMic_Click` is the sole consumer of `AudioDeviceManager.ToggleMute`, reached from the button click and the hotkey (via `DispatcherQueue.TryEnqueue`); both were and remain on the UI thread. The visuals/status/beep logic is byte-for-byte preserved; the only behavioral change is the intended coalescing of overlapping presses.

### Minimality
All changes are necessary and scoped to the mute toggle. Tabs respected, no unrelated refactors or formatting churn.

### Fixes applied

No issues remaining.

---
Reviewed at b84ab8ed322ccb95da171f0c043ee1f1005038e0
🤖 Reviewed with Claude Code

PR #198, opened 2026-07-04
Debounce slider settings persistence to stop per-tick file churn (#172)

jfbosch:
## What this changes

The three sliders in the main window — microphone level, text-to-speech
rate, and text-to-speech volume — each wrote the settings file on **every**
value change. A WinUI slider raises a change event for every intermediate
value during a pointer drag or a held arrow key (the main way a keyboard user
adjusts a slider), and each of those changes serialized the entire settings
object and atomically replaced the settings file — plus its `.bak` backup — on
the UI thread. Dragging or key-repeating a slider therefore triggered dozens
of full serializations and file replaces per second, causing audible stutter
in the slider feedback, disk churn, and needless backup rewrites.

This change defers (debounces) the file write. Each slider still updates the
in-memory settings immediately, but the write to disk now happens once, about
half a second after the last change, instead of on every tick. This mirrors a
pattern already used for the prompt text box.

## How it works

- A small, reusable `Debouncer` (in `Mutation.Ui/Core`) collapses a burst of
  rapid triggers into a single deferred action once the triggers stop for the
  configured quiet period. Its timing mechanism is injectable so its behaviour
  can be unit-tested without a running UI.
- Each of the three slider handlers now calls `Trigger()` instead of saving
  directly. The microphone slider keeps its live per-tick hardware level write,
  so there is no change to how quickly the mic responds — only the settings
  save is deferred.
- No data is lost on shutdown: the window's close handler already saves the
  settings unconditionally, and the in-memory value is up to date the instant
  the slider moves, so a still-pending debounced write is captured on close.

## Scope

Only the continuous slider streams are debounced. Discrete controls (toggles,
combo boxes) keep saving immediately. The two existing inline debounces
(prompt text, format) are left as-is.

## Test plan

- `dotnet test --configuration Release` — all 629 tests pass (6 new).
- New `DebouncerTests` cover: a rapid burst of triggers runs the action only
  once, a trigger after a completed run fires again, disposing while a run is
  pending cancels it, triggering after dispose is a no-op, and the constructor
  guards (null action, negative delay).

Manual (behavioural, unchanged persistence):
- Drag or hold an arrow key on the mic-level, TTS-rate, and TTS-volume sliders;
  confirm the value still persists across an app restart and that slider
  feedback is smooth.

Closes #172

### DIFF
--- a/Mutation.Tests/DebouncerTests.cs
+++ b/Mutation.Tests/DebouncerTests.cs
@@ -1,0 +1,136 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Mutation.Ui.Core;
+
+namespace Mutation.Tests;
+
+public class DebouncerTests
+{
+	// A delay stand-in the test drives by hand: each call records a pending "gate"
+	// the test completes explicitly (as if the time elapsed), so the coalescing and
+	// cancellation behaviour is exercised deterministically with no real time passing.
+	private sealed class ManualDelay
+	{
+		private readonly List<(TaskCompletionSource Tcs, CancellationToken Token)> _pending = new();
+		private readonly object _lock = new();
+
+		public Func<TimeSpan, CancellationToken, Task> Func => (_, token) =>
+		{
+			var tcs = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
+			lock (_lock)
+				_pending.Add((tcs, token));
+			token.Register(() => tcs.TrySetCanceled(token));
+			return tcs.Task;
+		};
+
+		public int Count
+		{
+			get { lock (_lock) return _pending.Count; }
+		}
+
+		// Complete the most recently scheduled delay as if its quiet period elapsed.
+		public void CompleteLast()
+		{
+			TaskCompletionSource tcs;
+			lock (_lock)
+				tcs = _pending[_pending.Count - 1].Tcs;
+			tcs.TrySetResult();
+		}
+	}
+
+	[Fact]
+	public void Trigger_RapidBurst_RunsActionOnce()
+	{
+		var clock = new ManualDelay();
+		int runs = 0;
+		using var ran = new ManualResetEventSlim(false);
+		using var debouncer = new Debouncer(
+			TimeSpan.FromMilliseconds(500),
+			() => { Interlocked.Increment(ref runs); ran.Set(); },
+			clock.Func);
+
+		debouncer.Trigger();
+		debouncer.Trigger();
+		debouncer.Trigger();
+
+		// Three triggers scheduled three delays; the first two were superseded and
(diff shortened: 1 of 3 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR debounces the three main-window slider saves (mic level, TTS rate, TTS volume) so a drag or held arrow key coalesces into a single settings write ~500ms after the last change, instead of a full serialize-and-atomic-replace per tick. It does this correctly and with good test coverage.

### Non-compliance
None. Build is green and `dotnet test --configuration Release` passes (629 tests, 0 failed). No accessibility regression — no UI/automation surface changed, only when the file write fires. No hardcoded-default gate concern: the debounce delay is an internal implementation detail of an existing behavior, not new user-facing configuration.

### Story alignment
Matches issue #172. Only the continuous slider streams are debounced; the mic slider keeps its live per-tick COM level write, and discrete controls (toggles, combo boxes) still save immediately. Scope is exactly as described.

### Correctness
- No data loss on shutdown: each handler mutates `_settings` in memory immediately and only the file write is deferred, and `MainWindow_Closed` (`MainWindow.xaml.cs:556`) saves `_settings` unconditionally before disposing the debouncer (`:2388`). A still-pending debounced write is therefore captured on close.
- UI affinity is preserved: `Debouncer.RunAsync` resumes with `ConfigureAwait(true)`, so `_action()` (the save) runs back on the UI thread that raised `ValueChanged`, exactly as the old inline save did.
- Trigger/Dispose are serialized under a lock; rapid triggers cancel and supersede the prior run, and `Trigger()` after `Dispose()` is a guarded no-op.

Minor observation (non-blocking, no change requested): the deferred `SaveSettingsToFile` runs fire-and-forget, so an I/O failure faults an unobserved Task rather than surfacing to the user. This is consistent with — and marginally safer than — the existing `TxtSpeechToTextPrompt_TextChanged` prompt debounce (`:1990`), which swallows the same way via `async void`. Not a regression; worth keeping in mind if settings-save error surfacing is ever generalized.

### Tests
`DebouncerTests` cover the meaningful paths with a deterministic injectable delay: rapid burst runs the action once, a trigger after a completed run fires again, dispose cancels a pending run, trigger-after-dispose is a no-op, and both constructor guards (null action, negative delay). Good coverage for the new unit.

### Regressions
None found. The mic slider's live hardware level write is untouched; only its settings persistence is deferred. Other `SaveSettingsToFile` call sites (combo boxes, prompt, close handler) are unchanged.

### Minimality
Tight and on-scope. The `Debouncer` is a small reusable unit in `Mutation.Ui/Core`, and only the three slider handlers plus the constructor and `Dispose` changed. No unrelated refactors.

### Fixes applied

No issues remaining.

---
Reviewed at 3764a4bae9ed7e60c7802e6377a4adf7af477e77
🤖 Reviewed with Claude Code

PR #199, opened 2026-07-04
Defer transcription CTS disposal so cancel can't race retries (#179)

jfbosch:
Closes #179

## What changed

Cancelling an in-progress transcription could surface a raw
`ObjectDisposedException` error dialog instead of a clean "Transcription
cancelled" message.

The cause: `SpeechToTextState.StopTranscription()` cancelled **and then
immediately disposed** the `CancellationTokenSource` that ties the
transcription to its cancellation. That method ran from two places — the UI
cancel path and the owning transcription task's cleanup. The OpenAI
speech-to-text service retries on failure, and each retry attempt links a
fresh cancellation token to that same source. When a user cancel landed
between two retry attempts, it disposed the source while the next attempt
was about to link to it, throwing `ObjectDisposedException`. That is not a
cancellation exception, so the layer that turns a cancel into a friendly
"Transcription cancelled" message did not recognise it and showed a raw
error dialog instead.

The fix splits the one "stop" operation into two, so cancelling never
disposes a source that retries might still use:

- **`CancelTranscription`** — signals cancellation only, no disposal. This
  is what the UI cancel path now calls. Retries still holding the token see
  the cancel and unwind cleanly as a cancellation.
- **`EndTranscription`** — cancels, disposes, and clears the source. Only
  the owning transcription task calls it, in its `finally`, once the awaited
  call and all of its retries have finished, so nothing can link to the
  source any more.

`StartTranscription` now also cancels a lingering prior source before
disposing it, rather than dropping it mid-flight.

One deliberate behaviour change: the app now reports a transcription as
"transcribing" from the moment of cancel until the task finishes tearing
down (a brief window) rather than flipping the state the instant cancel is
pressed. This is safer — the recorder lock is still held during teardown,
so the state now matches reality.

## Files

- `Mutation.Ui/Core/SpeechToTextState.cs` — split `StopTranscription` into
  `CancelTranscription` (signal only) and `EndTranscription` (dispose);
  cancel-before-dispose in `StartTranscription`.
- `Mutation.Ui/Core/SpeechToTextManager.cs` — the two transcription paths
  call `EndTranscription` in their `finally`; the public
  `CancelTranscription` delegates to the signal-only method.
- `Mutation.Tests/TranscriptionCancellationTests.cs` — new tests.

## Test plan

- `dotnet test --configuration Release` — all 645 tests pass, including 4
  new ones:
  - cancel signals cancellation without disposing the source (a fresh linked
    token can still be created and is already cancelled — no
    `ObjectDisposedException`);
  - end disposes the source and clears the transcribing flag, and is
    idempotent;
  - starting a new transcription cancels a lingering prior source;
  - end-to-end: a cancel while the service is mid-retry surfaces an
    `OperationCanceledException`, not an `ObjectDisposedException`.

### DIFF
--- a/Mutation.Tests/TranscriptionCancellationTests.cs
+++ b/Mutation.Tests/TranscriptionCancellationTests.cs
@@ -1,0 +1,165 @@
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using CognitiveSupport;
+using Mutation.Ui;
+
+namespace Mutation.Tests;
+
+// Covers issue #179: cancelling a transcription must not dispose the
+// CancellationTokenSource while background retries still link fresh tokens to
+// it. Disposal there surfaced an ObjectDisposedException as a raw error dialog
+// instead of a clean "Transcription cancelled".
+public class TranscriptionCancellationTests : IDisposable
+{
+	private readonly string _tempDirectory;
+	private readonly Settings _settings;
+
+	public TranscriptionCancellationTests()
+	{
+		_tempDirectory = Path.Combine(Path.GetTempPath(), "MutationTests_" + Guid.NewGuid().ToString("N"));
+		Directory.CreateDirectory(_tempDirectory);
+		_settings = new Settings
+		{
+			SpeechToTextSettings = new SpeechToTextSettings
+			{
+				TempDirectory = _tempDirectory
+			}
+		};
+	}
+
+	public void Dispose()
+	{
+		try
+		{
+			Directory.Delete(_tempDirectory, recursive: true);
+		}
+		catch (IOException)
+		{
+		}
+		catch (UnauthorizedAccessException)
+		{
+		}
+	}
+
+	[Fact]
+	public void CancelTranscription_SignalsCancellation_WithoutDisposingSource()
+	{
+		using var state = new SpeechToTextState(() => null);
+		var token = state.StartTranscription();
+
+		state.CancelTranscription();
+
+		Assert.True(token.IsCancellationRequested);
+
+		// A retry attempt links a fresh source to the token on every try. With the
+		// source only cancelled (not disposed) this must succeed and yield an
+		// already-cancelled token, never throw ObjectDisposedException.
+		using var thisTry = new CancellationTokenSource();
+		var ex = Record.Exception(() =>
+		{
+			using var linked = CancellationTokenSource.CreateLinkedTokenSource(token, thisTry.Token);
+			Assert.True(linked.IsCancellationRequested);
+		});
+		Assert.Null(ex);
+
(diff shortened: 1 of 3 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR fixes the `ObjectDisposedException`-on-cancel race in #179 by splitting the old `StopTranscription` into a signal-only `CancelTranscription` (used by the UI cancel path) and a `CancelTranscription`+dispose `EndTranscription` (called only from the owning transcription task's `finally`). It does exactly that, correctly.

### Non-compliance
None. `dotnet build --configuration Release` is clean (0 warnings), `dotnet test --configuration Release` is green (645 passed, 0 failed). No UI markup changed, no hidden defaults, no secrets.

### Story alignment
Matches issue #179 precisely: disposal is deferred until every retry has finished, and `StartTranscription` now cancels a lingering prior source before disposing it rather than abandoning it mid-flight. Both remedies the issue asked for are present.

### Correctness
- The disposed-source race is genuinely closed: `CognitiveSupport/OpenAiSpeechToTextService.cs:60` re-links `overallToken` (the caller's `_cts.Token`) on every retry attempt, so disposing the source between attempts is what threw. `EndTranscription` now runs only after `service.ConvertAudioToText(...)` and all its Polly retries complete (`SpeechToTextManager.cs:251`, `:290`), so nothing links to a disposed source.
- `EndTranscription` is correctly idempotent — it grabs and nulls `_cts` under `_ctsLock` before disposing, so the task `finally` and `Dispose()` can both run safely.
- Both UI cancel sites (`AudioSessionManager.cs:151`, `:420`) call the signal-only `CancelTranscription`, which is the right choice — they can run concurrently with the transcription still holding the recorder lock and retrying.
- The deliberate change to keep `TranscribingAudio` true through teardown is safe and, as noted, matches reality since the recorder lock is still held.

### Tests
Strong coverage. The four new tests exercise signal-only cancel (fresh linked token succeeds and is already cancelled, no `ObjectDisposedException`), dispose+clear+idempotency, lingering-prior-source cancellation, and an end-to-end race where a cancel mid-retry surfaces `OperationCanceledException` rather than `ObjectDisposedException`. The `RetryingCancellableService` fake faithfully mirrors the real service's per-attempt `CreateLinkedTokenSource` pattern.

### Regressions
None found. Every caller of the removed `StopTranscription` was updated; a repo-wide search shows no stale references. The signal-only vs. dispose split is applied consistently across both transcription paths.

### Minimality
All three changed files are necessary and scoped to the fix. No unrelated refactors or formatting churn.

### Fixes applied

No issues remaining.

---
Reviewed at a1b48e23147fc9fa7b3b7f3e3505b2cff4a265ce
🤖 Reviewed with Claude Code

PR #201, opened 2026-07-04
Move mic-level slider and record-start COM writes off the UI thread (#197)

### LINKED ISSUE #197: [DEBT] Mic-level slider and record-start write COM levels synchronously on the UI thread
# Context

Story #171 (PR #196) moved the microphone **mute toggle** off the UI thread so
its audio-device COM writes and failure-path device re-enumeration no longer
freeze the UI. Two sibling paths still run the same pattern synchronously on
the UI thread and were deliberately left out of that story because they need a
different, ordering-aware fix:

- The **microphone-level slider** — `SldMicLevel_ValueChanged`.
- The **record-start** path that re-asserts the pinned capture level.

Both call `MicrophoneLevelPinService.WriteLevel`, which writes the level over
the Windows audio COM API, reads it back to confirm, and on failure re-acquires
a fresh device reference (`RefreshEndpoint`) and retries — all on the UI thread.

# Evidence

- `Mutation.Ui/Core/MicrophoneLevelPinService.cs` — `WriteLevel` does the
  write + read-back verify, and on failure calls `_provider.RefreshEndpoint()`
  (a fresh device enumeration) before retrying.
- `Mutation.Ui/MainWindow.xaml.cs` — `SldMicLevel_ValueChanged` calls
  `_micLevelPinService.ApplyLevel(...)` on every slider tick; the record-start
  path calls `ReassertPinnedLevel(...)`.

# Impact

On a healthy device the write is a few milliseconds, but the failure/retry
path re-enumerates devices on the UI thread and can briefly freeze the UI and
the screen reader — the same problem #171 fixed for mute. The slider makes it
worse: `ValueChanged` fires many times per drag or per held arrow key, so any
stall repeats rapidly.

# Suggested fix

Move these writes off the UI thread, but — unlike the mute toggle — with an
**ordering-safe** design. A slider drag produces a burst of level requests, so
a naive per-tick `Task.Run` could apply writes out of order and settle on the
wrong level. Use a coalesce-to-latest background worker: always apply the most
recently requested level and drop superseded ones, marshaling only the final
outcome back to the UI. Keep device re-enumeration off the UI thread.

`MicrophoneMuteToggleCoordinator` (added in #171) is the reference for the
off-thread + single-flight approach; this needs the coalesce-to-latest variant
rather than drop-if-busy.

# Notes

Does not block any current story. Discovered while implementing #171 (PR #196).

jfbosch:
Closes #197

## What this changes

Story #171 moved the microphone **mute toggle** off the UI thread. Two sibling paths that write the microphone **capture level** were left running synchronously on the UI thread because they needed a different, ordering-aware fix:

- the **input-level slider** (fires many times per drag or per held arrow key), and
- the **record-start** re-assert of the pinned level.

Both eventually call the pin service, which writes the level over the Windows audio COM API, reads it back to confirm, and — on failure — re-enumerates the audio device and retries. On the UI thread that failure path can briefly freeze the window and the screen reader, and the slider makes it worse by repeating rapidly.

This change moves those writes off the UI thread with a design that is safe for a burst of rapid requests.

## How it works

A new `MicrophoneLevelWriteCoordinator` (in `Mutation.Ui/Core`) owns a single background worker that serializes every capture-level write. Its behavior differs from the mute coordinator in one important way:

- The mute coordinator **drops** an overlapping toggle (a second press while one is running).
- This coordinator **coalesces to the latest**: while a write is in flight, newer requests replace the queued one, and only the most recent survives. A slider drag therefore applies its first and its final value and skips the stale in-between ones, so the device settles on exactly what the user chose without a backlog of out-of-order writes. Superseded requests complete as "dropped" and never touch the device, so a whole drag can raise at most one failure warning.

Because a **single shared instance** (registered in DI) is used by every writer — the slider, the pin toggle, and the record-start re-assert — it is also the one place the level COM endpoint is written, so two threads never write it at once.

Call sites updated:

- **Slider** (`SldMicLevel_ValueChanged`) and **pin toggle** (`TglPinMicLevel_Toggled`) in `MainWindow` now hand their write to the coordinator instead of calling the pin service on the UI thread. The pin value is still persisted synchronously, so a coalesced/dropped write never loses the user's choice; only the file write stays debounced (issue #172).
- **Record-start** (`AudioSessionManager`) now awaits the pinned-level re-assert on the shared worker before the recorder starts, so the level is settled before capture begins and the failure-path re-enumeration no longer runs on the UI thread. The failure signal (beep + status line) is unchanged.

The one-time startup / mic-change re-assert stays synchronous — it is a single, non-bursty write that runs before any slider interaction, so it carries no concurrent-write hazard. It can be moved later if that startup path ever needs it.

## Scope note

The first commit on this branch is a WIP snapshot: a concurrent session working a different story found this work uncommitted in a shared checkout and preserved it to this branch. The second commit completes it. The combined branch is the full feature.

## Test plan

- `dotnet test --configuration Release` — green (655 passed, 0 failed).
- New unit tests for `MicrophoneLevelWriteCoordinator`:
  - the write runs off the calling thread,
  - coalesce-to-latest applies the first and last requests and drops the middle ones, in order,
  - a throwing write is surfaced as `Failed` and the worker recovers,
  - a new write is accepted after the previous one completes.

### Manual (needs a machine with a controllable-level mic)

1. Drag the input-level slider quickly / hold an arrow key — the UI stays responsive and the device ends at the released value.
2. With the pin on, start recording — the pinned level is re-asserted and "Listening" appears; the UI does not stall even if the device is momentarily busy.
3. Toggle the pin on — the current slider level is applied without freezing the window.

### DIFF
--- a/Mutation.Tests/MicrophoneLevelWriteCoordinatorTests.cs
+++ b/Mutation.Tests/MicrophoneLevelWriteCoordinatorTests.cs
@@ -1,0 +1,159 @@
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Mutation.Ui.Core;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class MicrophoneLevelWriteCoordinatorTests
+{
+	[Fact]
+	public void Constructor_NullWrite_Throws()
+	{
+		Assert.Throws<ArgumentNullException>(() => new MicrophoneLevelWriteCoordinator(null!));
+	}
+
+	[Fact]
+	public async Task RequestLatestAsync_ReturnsTheWriteResult()
+	{
+		var coordinator = new MicrophoneLevelWriteCoordinator(
+			_ => new CaptureLevelResult(CaptureLevelOutcome.Failed));
+
+		var result = await coordinator.RequestLatestAsync(30);
+
+		Assert.Equal(new CaptureLevelResult(CaptureLevelOutcome.Failed), result);
+	}
+
+	[Fact]
+	public async Task RequestLatestAsync_RunsTheWriteOffTheCallingThread()
+	{
+		// The write signals it started, then blocks. If it ran synchronously on this
+		// thread, RequestLatestAsync would block inside it and never hand back a task —
+		// so reaching the asserts below, with "started" already signalled, proves the
+		// write runs on another thread.
+		var started = new ManualResetEventSlim(false);
+		var release = new ManualResetEventSlim(false);
+		var coordinator = new MicrophoneLevelWriteCoordinator(_ =>
+		{
+			started.Set();
+			release.Wait();
+			return new CaptureLevelResult(CaptureLevelOutcome.Applied);
+		});
+
+		var task = coordinator.RequestLatestAsync(50);
+
+		Assert.True(started.Wait(TimeSpan.FromSeconds(5)), "write did not start on a background thread");
+		Assert.False(task.IsCompleted);
+		Assert.True(coordinator.IsWriting);
+
+		release.Set();
+		var result = await task;
+
+		Assert.Equal(new CaptureLevelResult(CaptureLevelOutcome.Applied), result);
+		AssertEventually(() => !coordinator.IsWriting, "worker did not settle after the write");
+	}
+
+	[Fact]
+	public async Task RequestLatestAsync_CoalescesToLatest_AppliesFirstAndLast_DropsMiddle()
+	{
+		var started = new ManualResetEventSlim(false);
+		var release = new ManualResetEventSlim(false);
+		var applied = new List<int>();
+		bool firstCall = true;
+
+		CaptureLevelResult Write(int level)
+		{
+			bool isFirst;
+			lock (applied)
+			{
+				isFirst = firstCall;
+				firstCall = false;
+				applied.Add(level);
+			}
+
+			// Block only the first write, so the burst below queues behind it.
+			if (isFirst)
+			{
+				started.Set();
+				release.Wait();
+			}
+
+			return new CaptureLevelResult(CaptureLevelOutcome.Applied);
+		}
+
+		var coordinator = new MicrophoneLevelWriteCoordinator(Write);
+
+		var first = coordinator.RequestLatestAsync(10);
+		Assert.True(started.Wait(TimeSpan.FromSeconds(5)), "first write did not start");
+
+		// While the first write is blocked, three more requests queue. Each supersedes
+		// the previous still-queued one, so only the last survives to be written.
+		var superseded1 = coordinator.RequestLatestAsync(20);
+		var superseded2 = coordinator.RequestLatestAsync(30);
+		var latest = coordinator.RequestLatestAsync(40);
+
+		// The two middle requests were dropped before they ran.
+		Assert.Null(await superseded1);
+		Assert.Null(await superseded2);
+
+		release.Set();
+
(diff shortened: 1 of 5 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR moves the microphone capture-level writes for the input-level slider and the record-start pinned-level re-assert off the UI thread, behind a new `MicrophoneLevelWriteCoordinator` that serializes writes on a single background worker and coalesces a burst (slider drag / held arrow key) to its most recent value. It does this correctly and matches the ordering-safe design called for in #197.

### Non-compliance
None. `dotnet build --configuration Release` is clean (0 warnings, 0 errors) and `dotnet test --configuration Release` is green (655 passed, 0 failed). No accessibility regression — the change's whole purpose is to keep the COM write and its failure-path device re-enumeration off the UI thread so the screen reader never freezes. No hidden defaults, no secrets.

### Story alignment
Implements exactly what the issue asks: an off-thread, coalesce-to-latest worker for the slider and record-start paths, keeping device re-enumeration off the UI thread. The startup / mic-change re-assert is deliberately left synchronous, which the issue explicitly scoped out.

### Correctness
The coordinator's single-flight logic is sound. `_workerRunning` and the pending slot are both guarded by `_gate`, and the worker's "no pending → stop" decision and `RequestLatestAsync`'s "not running → start" decision are serialized under the same lock, so there is no lost-wakeup and no double-worker window. Every `TaskCompletionSource` is completed on exactly one of two paths — superseded (`null`) or picked up by the drain (`result`) — so no awaiter can hang. A throwing write is caught in `Drain` and surfaced as `Failed` without killing the worker (`MicrophoneLevelWriteCoordinator.cs:120-134`). The in-flight write is never dropped: once the worker clears `_pendingCompletion`, a newer request supersedes only the *queued* slot, so the running write completes with its real result.

Call sites are correct: `SldMicLevel_ValueChanged` and `TglPinMicLevel_Toggled` record the pin intent synchronously on the UI thread before awaiting, so a coalesced/dropped COM write never loses the user's stored choice (`MainWindow.xaml.cs:1378-1394`, `:1357-1362`). `AudioSessionManager.ReassertPinnedLevelOffThreadAsync` treats a superseded (`null`) result as `Unchanged`, which is right — the device settled on the newer value (`AudioSessionManager.cs:235-242`).

### Tests
Good coverage for the new type: off-thread execution, coalesce-to-latest (first + last applied, middle dropped, in order), throwing-write-surfaces-as-Failed with worker recovery, and accept-after-complete. The `AssertEventually` helper avoids the known race with the worker clearing its running flag one loop pass after signalling.

### Regressions
Callers of the pin service are unchanged in behavior; the record-start re-assert now awaits the shared worker but preserves the same failure signal (beep + status). One minor observation, not blocking and not a functional regression: because the startup / mic-change re-assert at `MainWindow.xaml.cs:1322` still writes the level synchronously (bypassing the coordinator), it is now the one path that can, in principle, write the COM endpoint concurrently with the background worker (previously everything was UI-thread-serialized). The window is tiny (it requires a mic change while a slider write is still draining), the pin service's write is try/guarded with read-back verify + retry, and the author documents this as a deliberate scope decision that can be routed through the coordinator later. Acceptable as-is.

### Minimality
All changes are necessary and on-topic for the story. No unrelated refactors or formatting churn.

### Fixes applied

No issues remaining.

---
Reviewed at d9786b5145d257d9131d0d28444af56d68463b86
🤖 Reviewed with Claude Code

PR #205, opened 2026-07-05
Skip idle microphone waveform redraws (#173)

jfbosch:
## What this changes

The microphone waveform visualization redrew itself on a 33 ms timer for the
application's entire lifetime. Every tick copied the capture buffer, scanned the
sample window to compute RMS and peak, ran a full ScottPlot refresh, and updated the
level bar and pulse overlay — and it did all of that even when no microphone was being
captured or the buffer had not changed since the previous frame. That was constant
~30-times-per-second work on the UI thread (with the CPU and battery cost that implies),
sharing the same dispatcher queue that also delivers hotkey presses.

This PR makes the redraw conditional on whether new audio has actually arrived since the
last frame.

## How it works

A small, thread-safe latch (`WaveformRenderGate`) sits between the audio capture callback
and the render tick:

- The capture callback marks the gate whenever new samples land in the ring buffer.
- The render tick asks the gate whether anything new arrived. If not, it returns
  immediately — skipping the buffer copy, the RMS/peak scan, the plot refresh, and the
  meter writes — leaving the plot, level bar, and pulse overlay showing their last state.

During active capture the audio backend delivers buffers faster than the tick fires
(roughly every 15 ms versus the 33 ms tick), so the gate is always marked and the visible
behavior is unchanged. The saving is in the idle state — no microphone resolved, capture
failed to start, or capture stopped — where each tick now costs a single flag read instead
of a full render.

The gate is its own class so the "should this frame redraw?" decision is unit-testable on
its own, matching the existing pattern of small focused helpers in `Mutation.Ui.Core`.

## Acceptance criteria addressed

- Skips the refresh and meter updates when no capture is active or the buffer is unchanged
  (the issue's suggested fix). The timer is intentionally left running and the tick simply
  early-returns, which removes the idle churn with the least behavioral risk.

## Test plan

- `dotnet test --configuration Release` — all 660 tests pass, including 5 new tests for the
  render gate (no data → no render; one render per data arrival; multiple arrivals coalesce
  into a single render; cross-thread marking is observed by the consumer).
- The gate logic is covered by unit tests; the controller wiring (capture callback marks the
  gate, tick consumes it) is verified by inspection since it depends on WinUI/NAudio runtime
  types that are not exercised in the headless test host.

## Notes

- The only visible difference is in the idle state: after capture stops, the pulse overlay
  now freezes at its last value rather than fading a few more sub-perceptible frames.
  Active-capture behavior is identical.

Closes #173

### DIFF
--- a/Mutation.Tests/WaveformRenderGateTests.cs
+++ b/Mutation.Tests/WaveformRenderGateTests.cs
@@ -1,0 +1,63 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Mutation.Ui.Core;
+using Xunit;
+
+namespace Mutation.Tests;
+
+public class WaveformRenderGateTests
+{
+	[Fact]
+	public void ConsumeShouldRender_WithoutData_ReturnsFalse()
+	{
+		var gate = new WaveformRenderGate();
+
+		Assert.False(gate.ConsumeShouldRender());
+	}
+
+	[Fact]
+	public void ConsumeShouldRender_AfterDataArrived_ReturnsTrueThenFalse()
+	{
+		var gate = new WaveformRenderGate();
+		gate.MarkDataArrived();
+
+		Assert.True(gate.ConsumeShouldRender());
+		Assert.False(gate.ConsumeShouldRender());
+	}
+
+	[Fact]
+	public void ConsumeShouldRender_CoalescesMultipleArrivalsIntoOneRender()
+	{
+		var gate = new WaveformRenderGate();
+		gate.MarkDataArrived();
+		gate.MarkDataArrived();
+		gate.MarkDataArrived();
+
+		Assert.True(gate.ConsumeShouldRender());
+		Assert.False(gate.ConsumeShouldRender());
+	}
+
+	[Fact]
+	public void ConsumeShouldRender_ReRendersAfterEachNewArrival()
+	{
+		var gate = new WaveformRenderGate();
+
+		gate.MarkDataArrived();
+		Assert.True(gate.ConsumeShouldRender());
+		Assert.False(gate.ConsumeShouldRender());
+
+		gate.MarkDataArrived();
+		Assert.True(gate.ConsumeShouldRender());
+		Assert.False(gate.ConsumeShouldRender());
+	}
+
+	[Fact]
+	public async Task MarkDataArrived_FromAnotherThread_IsObservedByConsumer()
+	{
+		var gate = new WaveformRenderGate();
+
+		await Task.Run(gate.MarkDataArrived);
+
+		Assert.True(gate.ConsumeShouldRender());
+	}
+}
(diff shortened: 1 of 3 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

The PR stops the microphone waveform from re-rendering ~30 times a second for the app's whole lifetime by gating each redraw on whether new audio actually arrived since the last frame. The logic is correct, race-safe, and does not change what the user sees during active capture.

### Non-compliance
None. The solution builds and `dotnet test --configuration Release` is green (660 tests, including 5 new). No UI/accessibility surface changed — this only lowers redraw frequency, so there are no `AutomationProperties`, labeling, focus, or keyboard-trap implications. No hidden hardcoded defaults and no secrets.

### Story alignment
Matches issue #173 exactly. The issue's suggested fix was to skip `Refresh()` and meter updates when no capture is active or the buffer is unchanged; that is what the render gate does. The optional stronger fix (stopping the timer entirely while idle) was deliberately not taken — the tick simply early-returns — which removes the churn with less behavioral risk. That trade-off is reasonable and documented in the PR body.

### Correctness
Traced the three edited sites against every caller:

- Startup (`MainWindow.xaml.cs:187` `Initialize()` then `:212` `StartCapture()`): `Initialize()` draws the flat baseline with its own `Refresh()`, and `StartCapture()` marks the gate — including on the `MicrophoneDeviceIndex < 0` early-return path (`MicrophoneVisualizationController.cs:116`) — so the first idle frame renders once and the plot can never get stuck blank.
- The gate is marked from the capture thread in `OnWaveformDataAvailable` *after* the buffer lock is released, and consumed on the UI thread, so there is no nested-lock or ordering hazard. The read-and-reset in `ConsumeShouldRender` is atomic under its own lock.
- Worst-case interleaving only ever delays a frame by one ~33 ms tick or causes one redundant redraw; no captured samples are ever dropped from rendering, and active capture (buffers every ~15 ms) self-corrects immediately.

### Tests
`WaveformRenderGate` — the piece carrying the actual decision — is fully covered: no-data yields no render, one render per data arrival, multiple arrivals coalesce into a single render, and a cross-thread mark is observed by the consumer. The controller wiring itself is not unit-tested because it depends on WinUI/NAudio runtime types the headless test host does not exercise; extracting the gate into `Mutation.Ui.Core` is the right call to make the decision testable in isolation, matching the existing coordinator pattern.

### Regressions
None. During active capture the audio backend delivers buffers faster than the tick fires, so the gate is always marked and behavior is identical to before. The only visible difference is in the idle state: after capture stops, the pulse overlay freezes at its last value instead of fading a few more sub-perceptible frames — an acceptable and disclosed change.

### Minimality
Every changed line serves the fix. No unrelated refactors, formatting churn, or bundled work. Tabs throughout, per repo convention.

### Fixes applied

No issues remaining.

---
Reviewed at 9af2e386c67469a33d8ee390b79f295906444371
🤖 Reviewed with Claude Code

PR #206, opened 2026-07-05
Stream PCM into fixed frames without per-byte buffering (#174)

jfbosch:
Closes #174

## What changed

Both audio paths that produce Opus frames — importing a file (`AudioFileConverter`) and live recording (`AudioRecorder`) — used to accumulate decoded 16-bit PCM one element at a time into a growing `List`, then pull each 20 ms (960-sample) frame out of the front with `GetRange(...).ToArray()` followed by `RemoveRange(...)`. `RemoveRange` shifts every remaining byte in the list forward on each frame, so the buffer (up to ~96 KB) was recopied once per frame. For a one-hour import that is on the order of hundreds of millions of single-element `Add` calls plus roughly 180,000 whole-buffer shift cycles, all of it running before transcription can even begin.

This replaces that with a small, reusable helper, `PcmFrameSplitter`, that turns a stream of little-endian PCM bytes into fixed-size `short[]` frames using one fixed carry buffer and `Buffer.BlockCopy`:

- Whole frames are copied straight out of the incoming buffer — no intermediate list, no per-byte adds, no per-frame buffer shifts.
- Bytes that do not complete a frame are held in a single carry buffer and combined with the next chunk.
- Each emitted frame is a freshly allocated array. This matters because `SilenceTrimmer` keeps references to frames it is handed; reusing one buffer would corrupt the trimmed output.

Both call sites keep their existing end-of-stream behavior:

- `AudioFileConverter` pads the final partial frame with silence and emits it (never drops trailing audio on import).
- `AudioRecorder` drops its trailing sub-20 ms tail, and now uses a fresh splitter per recording, which also stops a small partial-frame remainder from bleeding from one recording into the next.

Using little-endian `Buffer.BlockCopy` produces the same samples as the previous manual decode (`(short)((b[i*2+1] << 8) | b[i*2])`) on every platform this app supports (Windows x64/ARM64/ARM32, all little-endian), and `AudioFileConverter` already used `Buffer.BlockCopy`.

## Files

- `CognitiveSupport/PcmFrameSplitter.cs` — new helper (single responsibility, unit-tested in isolation).
- `CognitiveSupport/AudioFileConverter.cs` — use the splitter in place of the `List<byte>` accumulation loop.
- `CognitiveSupport/AudioRecorder.cs` — use a per-recording splitter in place of the `List<short>` buffer.
- `Mutation.Tests/PcmFrameSplitterTests.cs` — new tests.

## Test plan

- `dotnet test --configuration Release` — all 685 tests pass, including 16 new `PcmFrameSplitter` tests.
- New tests cover: whole-frame emission, partial frames completed across calls, carry-over remainder, byte-at-a-time feeding, honoring `count` over buffer length, flush with and without silence padding, empty flush, frame-array independence (the trimmer-retention guarantee), little-endian decode, and argument validation.
- The two call-site behaviors (pad-and-emit on import, drop-tail on record) are preserved; the existing recorder and speech-to-text tests continue to pass.

### DIFF
--- a/CognitiveSupport/AudioFileConverter.cs
+++ b/CognitiveSupport/AudioFileConverter.cs
@@ -72,11 +72,10 @@ public static class AudioFileConverter
 		byte[] buffer = new byte[outFormat.AverageBytesPerSecond];
 		int bytesRead;
 
-		// Buffer for accumulation to feed fixed frame sizes to Opus (960 samples = 1920 bytes)
-		// Opus requires 2.5, 5, 10, 20, 40, or 60ms frames. We use 20ms (960 samples).
+		// Split the decoded byte stream into fixed frame sizes for Opus (960 samples = 1920
+		// bytes). Opus requires 2.5, 5, 10, 20, 40, or 60ms frames. We use 20ms (960 samples).
 		int samplesPerFrame = 960;
-		int bytesPerFrame = samplesPerFrame * 2;
-		List<byte> accumulationBuffer = new List<byte>();
+		var splitter = new PcmFrameSplitter(samplesPerFrame);
 
 		var trimmer = silenceOptions is null
 			? null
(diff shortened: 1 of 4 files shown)


jfbosch:
## Review by Claude

**Verdict: Approved**

This PR replaces the per-byte `List` accumulation + `GetRange`/`RemoveRange` frame extraction in both audio paths with a single-responsibility `PcmFrameSplitter` that emits fixed-size `short[]` frames via `Buffer.BlockCopy` over a fixed carry buffer. It does what the issue asks and does it correctly.

### Non-compliance
None. Builds clean and `dotnet test --configuration Release` is green (685/685, including 16 new splitter tests).

### Story alignment
Matches issue #174 exactly. Both call sites named in the issue (`AudioFileConverter`, `AudioRecorder`) are converted to the carry-buffer + `Buffer.BlockCopy` approach the issue suggested, eliminating per-byte adds and per-frame O(n) list shifts. No out-of-scope work bundled in.

### Correctness
- `PcmFrameSplitter.Append` (`CognitiveSupport/PcmFrameSplitter.cs:50`) is sound: carry top-up consumes from `source[0]` before `offset` advances, whole frames copy straight from the source, and the sub-frame tail is stashed. Because `_carry` is reset to 0 whenever it fills, `_carryLength` is always `< _frameBytes` between calls, so the top-up path can never overflow.
- `ToFrame` allocates a fresh array per frame (`PcmFrameSplitter.cs:113`), preserving the `SilenceTrimmer` frame-retention guarantee — verified by the `EmittedFrames_AreIndependentArrays` test.
- Little-endian `Buffer.BlockCopy` produces identical samples to the previous manual `(short)((b[i*2+1] << 8) | b[i*2])` decode on the little-endian-only platforms this app targets.
- End-of-stream behavior is preserved: `AudioFileConverter` pads-and-emits the final partial frame (`AudioFileConverter.cs:99`) then flushes the trimmer in the correct order; `AudioRecorder` drops the sub-20ms tail by nulling the splitter in `StopRecording` (`AudioRecorder.cs:178`).
- Recorder error handling on the capture thread is consistent — the `_oggStream is null || _pcmFrameSplitter is null` guard matches the null-out on exception.

Minor, non-blocking (not worth a change): the new splitter carries an odd trailing byte across `Append` calls, whereas the old recorder discarded it via `BytesRecorded / 2`. For 16-bit mono capture `BytesRecorded` is always even, so this never triggers — and the new behavior is strictly more correct.

### Tests
Comprehensive. The 16 new tests cover whole-frame emission, cross-call completion, carry remainder, byte-at-a-time feeding, honoring `count` over buffer length, flush with/without padding, empty flush, frame independence, little-endian decode, and argument validation. Both call-site behaviors remain covered by the existing recorder/STT suites.

### Regressions
None found. `PcmFrameSplitter` has no consumers outside the two converted call sites, and both preserve their prior observable behavior.

### Minimality
Every changed line serves the refactor. No unrelated formatting or refactors.

### Fixes applied

No issues remaining.

---
Reviewed at 965aac606deb14bf0afcb73d1bf0f472b7b27c78
🤖 Reviewed with Claude Code

PR #291, opened 2026-08-07
Stop four quiet lies: welded words, a casing typo, a lost keyterm, a missing reason

### LINKED ISSUE #222: [BUG] Smart transcript rules crash on regex metacharacters and delete surrounding whitespace
## Context

`TextFormatter.FormatWithRule`, `MatchTypeEnum.Smart` branch (`CognitiveSupport/TextFormatter.cs:135-136`):

```csharp
string pattern = $@"(\b|^)([.,]?)([ ]*{find}[.,]?[ ]*)(\b|$)";
string replacement = $"$1$2{replaceWith}$4";
text = Regex.Replace(text, pattern, replacement, regexOptions);
```

Two distinct defects on these two lines.

### 1. Unescaped interpolation (crash)

`Smart` is documented — by contrast with the separate `RegEx` mode — as a literal match, but `find` goes straight into the pattern and `replaceWith` straight into the substitution.

A Smart rule `Find: "C++"` → `ReplaceWith: "C plus plus"` produces the pattern `…[ ]*C++[.,]?…`, which `Regex.Replace` rejects with `RegexParseException: Nested quantifier '+'`. Because this throws out of `FormatWithRules`, **every** transcript-formatting run fails from that point on — not just that one rule. Same for `(`, `[`, `\`. A `replaceWith` containing `$1` silently expands a capture group instead of inserting a literal.

### 2. Whitespace loss (corruption)

Group 3 swallows the spaces on either side of the match and the replacement never restores them. Input `"contain ai also paint"` with rule `ai` → `AI` yields `"containAIalso paint"` — adjacent words are welded together.

This is currently **pinned as correct** by `Mutation.Tests/TextFormatterTests.cs:69` (`FormatWithRule_Smart_OnlyMatchesAtWordBoundary`), whose own comment concedes "Pins down current behavior". The companion test at `:58` asserts only `Assert.Contains("AI", result)` and `Assert.DoesNotContain(" ai ", result)`, both of which hold for the mangled output.

## Requirements

- Escape the literal: `Regex.Escape(find)`, and escape `$` in the replacement (`replaceWith.Replace("$", "$$")`).
- Preserve whitespace: restructure to `(\b|^)([.,]?)([ ]*)({find})([.,]?)([ ]*)(\b|$)` and emit the captured spacing groups.
- **Update the two tests** that currently enshrine the whitespace loss: change `TextFormatterTests.cs:69` to expect `"contain AI also paint"`, and replace the weak `Contains`/`DoesNotContain` pair at `:58` with a full-string `Assert.Equal("Please AI is great.", result)`.
- Add cases for `C++`, `(`, `[`, and a `replaceWith` containing `$1`.

## Notes

Found during a full codebase audit; flagged independently by both the production-code and test-suite passes.

### LINKED ISSUE #240: [BUG] OpenAI and Anthropic services disagree on model-name lookup semantics
## Context

Two contract divergences behind the same `ILlmService` interface:

| | `LlmService` (OpenAI) | `AnthropicLlmService` |
| --- | --- | --- |
| Lookup | `new Dictionary<string, ChatClient>()` — ordinal, **case-sensitive** (`CognitiveSupport/LlmService.cs:34-48`) | `ToDictionary(..., StringComparer.OrdinalIgnoreCase)` (`CognitiveSupport/AnthropicLlmService.cs:47`) |
| Duplicate model names | indexer assignment silently keeps the **last** duplicate | `ToDictionary` **throws** `ArgumentException` at construction |

## Failure scenario

A user hand-edits `Mutation.json` with `"ModelName": "GPT-4.1"` against a configured model named `"gpt-4.1"`. `CompositeLlmService` routes on its own `_modelProviders` dictionary and, if that resolves, hands off to `LlmService`, which throws `ArgumentException: GPT-4.1 is not one of the configured models`.

The identical typo against an Anthropic model works fine. Same file, same kind of typo, two different outcomes — and the error message does not hint that casing is the problem.

## Requirements

- Use `StringComparer.OrdinalIgnoreCase` in `LlmService` to match `AnthropicLlmService`.
- Validate duplicate model names identically in both, with the same exception type and a message naming the duplicate.
- Add tests covering case-mismatched lookup and duplicate names for both providers.

## Notes

Found during a full codebase audit.

### LINKED ISSUE #245: [BUG] Deepgram service truncates keyterms at the first period and ignores the cancellation token
## Context

Two defects in `CognitiveSupport/DeepgramSpeechToTextService.cs`.

### 1. Keyterm parsing truncates at the first period (`:108`)

```csharp
@"(?<=\bkeyterms:\s*).*?(?=\.)"
```

Non-greedy up to the **first** `.`. The prompt `"keyterms: Dr. Bosch, Mutation, WinUI."` yields the single keyterm `"Dr"` — everything after the first abbreviation period is silently dropped. If the prompt has no trailing `.` at all, the match fails and **all** keyterms are dropped.

Keyterms exist to improve recognition of names and jargon, which is exactly the text most likely to contain abbreviation periods.

### 2. Cancellation token not passed to file I/O (`:41`)

```csharp
await File.ReadAllBytesAsync(audioffilePath)
```

The `overallCancellationToken` is available but not passed. A cancelled transcription still blocks reading a large imported file off disk.

## Requirements

- Anchor the keyterm regex to end-of-line (`(?=\.\s*$|$)`) or require an explicit terminator; add tests for abbreviation periods and a missing trailing period.
- Pass `overallCancellationToken` to `ReadAllBytesAsync`.

## Notes

Found during a full codebase audit. `DeepgramSpeechToTextService` has zero test coverage, while the interchangeable `OpenAiSpeechToTextService` has some — see the coverage issue.

jfbosch:
Four contained faults from the codebase audit, batched because they all live in
code that composes text a user reads or hears, and none is big enough to be worth
a branch of its own.

Closes #222, closes #240, closes #245, closes #288.

## #222 — Smart transcript rules

Two defects on the same two lines.

**Crash.** `Find` went straight into the pattern, so a Smart rule finding `C++`
compiled to a nested quantifier and threw — out of `FormatWithRules`, taking every
later rule with it, not just that one. `Find` is now `Regex.Escape`d.

`ReplaceWith` goes in through a `MatchEvaluator` rather than as a substitution
string, so `$1` is the two characters the user typed. That is stronger than
escaping `$`: nothing in the replacement is ever interpreted.

**Whitespace loss.** The pattern swallowed the spaces either side of the match and
never gave them back, welding `contain ai also` into `containAIalso`. The spacing
is now captured and re-emitted.

> **Departure from the issue.** The issue prescribed emitting the captured spacing
> unconditionally. That would have regressed every shipped default rule: they are
> dictated punctuation — `full stop` → `". "`, `comma` → `", "`, `new line` →
> newline — where the replacement supplies its own spacing and re-emitting the gap
> would give `Hello .  next word`. So the spacing is emitted only where the
> replacement does not already provide it. A replacement that is punctuation plus
> its spacing attaches to the preceding word; one that merely *starts* with
> punctuation and then carries text (`dotnet` → `.NET`) is a word and keeps its
> gap. Both cases are pinned by tests.

The two tests that enshrined the whitespace loss are updated as the issue asked,
and the weak `Contains`/`DoesNotContain` pair is now a full-string assertion.

## #240 — Model-name lookup parity

OpenAI keyed its lookup ordinally, Anthropic case-insensitively. `"GPT-4.1"`
against a configured `"gpt-4.1"` failed on one provider and worked on the other.
A repeated name was silently deduped by one (last wins) and rejected by the other
with a framework message that did not name the duplicate.

Both now build their lookup through `CognitiveSupport/LlmModelIndex.cs`, which is
the single answer to "what does a model name mean": case-insensitive, unique,
and one exception type with one message.

Also aligned: `AnthropicLlmService` now rejects an empty model set, which
`LlmService` already did.

`LlmModelLookupParityTests` drives both implementations through the same cases so
they cannot drift apart again.

## #245 — Deepgram keyterms and cancellation

The keyterm regex was non-greedy to the *first* period anywhere in the prompt, so
`keyterms: Dr. Bosch, Mutation.` yielded the single term `Dr` — and keyterms exist
for precisely the names and jargon most likely to contain an abbreviation period.
A prompt with no closing period matched nothing and dropped every term.

The list now runs to the end of its own line, with one optional closing full stop.
`ParseKeyterms` is `internal` so the eight new cases can drive it directly; the
class had no test coverage at all before.

`overallCancellationToken` now reaches `File.ReadAllBytesAsync`.

## #288 — Error dialog keeps the middle of the chain

The summary was the outer message plus the innermost one. The actionable sentence
is usually neither: a provider 401 wrapped as `Speech to text failed.` over a
socket error read as a dropped connection, and the user never learned their API
key was wrong.

Every distinct level now speaks. Duplicates are collapsed against everything
already said, not just the previous line, so the existing
`SameMessageBothLevels_IsNotRepeated` behaviour holds and a non-consecutive repeat
is caught too.

> **The human decision the issue flagged.** The cap is **four lines**
> (`ErrorDialogMessage.MaxChainMessages`). The dialog is read aloud, so this is a
> listening budget; four is about as much as stays a headline. Past the cap the
> log pointer covers the rest, as the issue suggested. Chains that deep are rare.

## User guide

Updated in this PR, per the project rule:

- `transcript-formatting.md` — Smart takes Find exactly as typed, so symbols are fine.
- `settings.md` — how to write a Deepgram `keyterms:` line (previously undocumented);
  model names ignore capitals and must be unique.
- `troubleshooting.md` — what an error message now contains, with the 401 example.

HTML regenerated and committed alongside.

## Verification

`dotnet test --configuration Release` — **1235 passed, 0 failed.**

### DIFF
--- a/CognitiveSupport/AnthropicLlmService.cs
+++ b/CognitiveSupport/AnthropicLlmService.cs
@@ -34,7 +34,8 @@ public class AnthropicLlmService : ILlmService
 		int retryCount = 3)
 	{
 		if (string.IsNullOrEmpty(apiKey)) throw new ArgumentNullException(nameof(apiKey));
-		if (models is null) throw new ArgumentNullException(nameof(models));
+
+		var modelList = LlmModelIndex.Validate(models, nameof(models));
 
 		_apiKey = apiKey;
 		_httpClient = httpClient ?? throw new ArgumentNullException(nameof(httpClient));
@@ -44,7 +45,7 @@ public class AnthropicLlmService : ILlmService
 		_httpClient.Timeout = Timeout.InfiniteTimeSpan;
 		_timeoutSeconds = timeoutSeconds > 0 ? timeoutSeconds : 60;
 		_retryCount = retryCount < 0 ? 0 : retryCount;
-		_modelConfigs = models.ToDictionary(m => m.Name, m => m, StringComparer.OrdinalIgnoreCase);
+		_modelConfigs = modelList.ToDictionary(m => m.Name, m => m, LlmModelIndex.NameComparer);
 	}
 
 	public async Task<string> CreateChatCompletion(
--- a/CognitiveSupport/DeepgramSpeechToTextService.cs
+++ b/CognitiveSupport/DeepgramSpeechToTextService.cs
@@ -38,7 +38,7 @@ public class DeepgramSpeechToTextService : ISpeechToTextService
 
 		List<string> keyterms = ParseKeyterms(speechToTextPrompt);
 
-		var audioBytes = await File.ReadAllBytesAsync(audioffilePath).ConfigureAwait(false);
+		var audioBytes = await File.ReadAllBytesAsync(audioffilePath, overallCancellationToken).ConfigureAwait(false);
 		const string AttemptKey = "Attempt";
 
 		var delay = Backoff.LinearBackoff(TimeSpan.FromMilliseconds(500), retryCount: 3, factor: 1);
@@ -100,13 +100,25 @@ public class DeepgramSpeechToTextService : ISpeechToTextService
 		return response;
 	}
 
-	private static List<string> ParseKeyterms(string speechToTextPrompt)
+	/// <summary>
+	/// Reads the comma-separated keyterms out of a prompt line beginning <c>keyterms:</c>.
+	/// <para>
+	/// The list runs to the end of its line, with one optional full stop allowed to close it.
+	/// It used to stop at the first period anywhere, which meant "keyterms: Dr. Bosch, Mutation."
+	/// yielded the single term "Dr" — and keyterms exist precisely to help with names and jargon,
+	/// the text most likely to contain an abbreviation period. A prompt with no closing period
+	/// dropped every term instead (issue #245).
+	/// </para>
+	/// </summary>
+	internal static List<string> ParseKeyterms(string speechToTextPrompt)
 	{
 		if (speechToTextPrompt == null)
 			speechToTextPrompt = string.Empty;
 
-		string pattern = @"(?<=\bkeyterms:\s*).*?(?=\.)";
-		Match match = Regex.Match(speechToTextPrompt, pattern, RegexOptions.IgnoreCase);
+		// Multiline `$` matches before the `\n` of a CRLF, so the `\r` has to be spelled out
+		// or the lazy match swallows it — and with it the closing period it was meant to trim.
+		string pattern = @"(?<=\bkeyterms:[ \t]*).*?(?=\.?[ \t]*\r?$)";
+		Match match = Regex.Match(speechToTextPrompt, pattern, RegexOptions.IgnoreCase | RegexOptions.Multiline);
 		if (match.Success)
 		{
 			string keytermsString = match.Value.Trim();
--- a/CognitiveSupport/LlmModelIndex.cs
+++ b/CognitiveSupport/LlmModelIndex.cs
@@ -1,0 +1,53 @@
+namespace CognitiveSupport;
+
+/// <summary>
+/// The one place that decides what a configured LLM model name means, so the OpenAI and
+/// Anthropic services behind <see cref="ILlmService"/> cannot answer that question
+/// differently.
+/// <para>
+/// They used to. OpenAI keyed its lookup ordinally and Anthropic case-insensitively, so
+/// hand-editing <c>Mutation.json</c> to <c>"GPT-4.1"</c> against a configured
+/// <c>"gpt-4.1"</c> failed while the identical typo against an Anthropic model worked.
+/// A repeated name diverged the same way: OpenAI silently kept the last copy, Anthropic
+/// threw an unhelpful framework message (issue #240).
+/// </para>
+/// </summary>
+public static class LlmModelIndex
+{
+	/// <summary>
+	/// Model names come from a hand-edited settings file, where a difference in casing is a
+	/// typo rather than a different model, so both providers match them case-insensitively.
+	/// </summary>
+	public static StringComparer NameComparer => StringComparer.OrdinalIgnoreCase;
+
+	/// <summary>
+	/// Materialises the configured models and rejects the two sets a lookup cannot represent:
+	/// no models at all, and a name that repeats once casing is ignored.
+	/// </summary>
+	/// <param name="models">The models as configured, in file order.</param>
+	/// <param name="parameterName">The caller's parameter name, for the thrown exception.</param>
+	/// <exception cref="ArgumentNullException"><paramref name="models"/> is null.</exception>
+	/// <exception cref="ArgumentException">The set is empty, or a name is configured twice.</exception>
+	public static List<LlmModelConfig> Validate(
+		IEnumerable<LlmModelConfig> models,
+		string parameterName)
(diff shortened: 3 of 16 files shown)


jfbosch:
## Independent review — done, with fixes pushed

Two reviewers, run without sight of each other. The general pass raised nothing above
its confidence bar. The adversarial pass built a differential fuzz harness that compiles
`git show main:CognitiveSupport/TextFormatter.cs` alongside the branch build and runs both
through the full `ApplyRules` pipeline, and it found two real problems in the Smart rule.
Both are fixed in 81616e6.

**1. Symbol replacements regressed.** `IsSentencePunctuation` covered only `. , ; : ! ?` —
exactly the shipped set — so a user rule `dash` → `-` fell to the word branch and got a gap
on both sides: `state dash of dash the dash art` became `state - of - the - art` where
`main` correctly gave `state-of-the-art`. Same for `slash`, `at sign`, `underscore`,
`and sign`.

**2. Consecutive deletions left a run of spaces.** `so um um I think` → `so  I think`, three
`um`s → three spaces. The guide has always promised deletion works "without leaving a double
space", and the fix only delivered it for a single occurrence.

### What changed

One predicate was doing three jobs. It is now an explicit three-way split:

- **Symbol** — no letter or digit anywhere in `ReplaceWith`. Drops the gap on both sides,
  which is byte-for-byte what this code did before it learned to keep spacing at all. Every
  one of the sixteen shipped rules is a symbol, so no existing rule can be made worse by this
  branch **by construction**, not by sampling.
- **Deletion** — closes to one space however many fillers were said. This needs state across
  the pass: consecutive fillers are adjacent matches, so the second has no leading gap left
  to keep, and punctuation the first carried has no word left to attach to.
- **Word** — keeps its gap, as before. `.NET` is a word, not punctuation.

Fixing (2) surfaced a third, which the re-check caught and is also fixed: a line opening
`um, um, I think` stranded a comma per filler, because each deletion re-emitted the
punctuation and `CleanLine` only strips the first. And a deletion no longer doubles a space
the following text already supplies (`I think um , next`).

### Verification

- `dotnet test --configuration Release` — **1251 passed, 0 failed** (was 1235).
- Differential fuzz vs `main`, 200,000 randomised dictation strings × both
  `manualPunctuation` arms, all sixteen seeded defaults: **0 differences**.
- 900,000 fuzz cases total across deletion-only, custom-rule, and a "is new worse than main"
  classifier: **0 exceptions on either side**, and the only remaining categories where the new
  code differs adversely from `main` are the two cosmetic ones filed below.

The fuzz also turned up a latent `main` bug this PR happens to fix: a digit-only
`ReplaceWith` collided with the old `$1$2{replaceWith}$4` substitution's group numbering, so
`two` → `2` on "I have two apples" leaked a literal `$2` into the transcript. 14,068 of
300,000 fuzz cases hit it.

### Filed, not fixed here

#293 — Smart rules cannot express a symbol that attaches on only one side (`%`, `#`, `(`).
Not a regression; the pre-#222 code produced the same output. Two smaller cosmetic residuals
are recorded in the comments there: a `ReplaceWith` that itself opens with whitespace doubles
the gap when two matches abut (374 / 300,000), and a deletion run can leave a space before a
following tab (101 / 200,000). All three want the same attaches-left / attaches-right
classification, so they belong in one piece of work rather than three.

PR #292, opened 2026-08-07
Stop the app announcing four things that were not true

jfbosch:
Closes #226
Closes #232
Closes #234
Closes #271

Four small defects around the dictation round trip, taken together because they
sit in the same three files. Each one had the app telling a blind user something
it had not actually done.

## A held hotkey auto-repeated (#226)

`MOD_NOREPEAT` was declared in `HotkeyManager` and referenced nowhere else, so
Windows re-delivered `WM_HOTKEY` at the keyboard auto-repeat rate for as long as
the combination was held. Holding the dictation shortcut a moment too long
started a recording, stopped and transcribed it, and started another — with a
`RecorderBusyException` failure beep per round. The mute hotkey flip-flopped the
same way.

Modifier composition moves out to `HotkeyModifiers.Compose`, which always sets
the flag and can be tested without a window handle or a P/Invoke.

## SendKeys and paste reported success when the injection failed (#232)

Both delivery branches did `_ = Task.Run(...)` and returned true immediately, and
`HotkeyManager.SendText` discarded `SendInput`'s return value. When the
foreground application runs with higher privileges, Windows drops the injected
input silently — so Mutation played the success beep and announced "Transcript
ready" for text that never arrived, with no way for the user to find out.

- `SendText` and `SendHotkey` now return whether Windows accepted the keystrokes.
- `TryInsertIntoActiveApplicationAsync` awaits the injection and returns a
  `TranscriptDeliveryOutcome` instead of a bool.
- A delivery failure gets its own beep and its own message, distinct from the
  clipboard one — the causes and the remedies are different, and blaming the
  clipboard for a privilege problem sends the user hunting for the wrong thing.

Awaiting keeps the work off the UI thread (it runs on the thread pool; the UI
thread returns to its message pump) and is bounded — `SendInput` is synchronous,
and the hotkey path waits at most `ModifierReleaseTimeoutMs` for the user to
release the chord.

One judgement call worth flagging: `SendHotkey` returns false when it had to use
the WinForms `SendKeys` fallback, because that path cannot confirm the target
took the keystrokes. The fallback still runs as a best effort; it just does not
get to claim success. For a blind user a false alarm is recoverable and a false
success is not.

## FinalizeTranscript could leave the transcript box read-only forever (#234)

The two statements that gave the user their transcript box back ran last, after
three awaits and two beeps, with no `try`/`finally`. Any throw on the delivery
path skipped both, and because the method is `async void` the exception went to
the global handler — so the box simply stopped accepting typing for the rest of
the session, with nothing said about why.

Restoration moves into a `finally` through the new `GuardedUiOperation`, and the
failure is announced with status and beep rather than left to the log. The guard
also survives its own failure report throwing, which matters here because the
reporting path is beeps and automation peers — exactly what tends to be failing.

## Stopping a recording could announce and beep "recording started" (#271)

`StartStopRecordingAsync` raised `StateChanged` on the stop path before the
recorder had stopped, and the handler — dispatched to the UI thread, so running
after the raiser had moved on — branched on `IsRecording`. With the recorder
semaphore contended it took the *start* branch on a stop: "Recording..." in the
transcript box, the button into its Stop state, and `BeepType.Start`.

Fixed at the source. `StateChanged` is now `EventHandler<RecordingActivity>` and
each raise site states the activity it is entering; the handler never re-reads
recorder flags. `RecordingUiPlanner` turns the activity into what the user sees
and hears, so the mapping is assertable without a window.

**Behaviour change beyond the reported bug**, called out deliberately: the retry
and upload paths also raised a state change before their transcription began, so
they said "Transcribing your recording..." and then showed the *idle* state.
They now declare `Transcribing` like the record path does, which disables the
record buttons for the duration and shows the placeholder. This is the same
defect the issue describes and the fix falls out of it, but it is a visible
change and the reviewer should agree with it.

The two `catch` blocks still read the recorder's flags, because a failure can
leave the session either way and its own flags are the best information
available there. That is commented in place.

## Tests

30 new tests across four files; the whole suite is 1274 passing on Release.

- `HotkeyModifiersTests` — `MOD_NOREPEAT` set for all 16 modifier combinations,
  requested modifiers preserved, unrequested ones absent.
- `TranscriptDeliveryMessagesTests` — an injection failure is reported, does not
  mention the clipboard, wins over a simultaneous clipboard failure, and every
  failure ends by saying where the text still is.
- `GuardedUiOperationTests` — state restored when the work throws synchronously,
  when it throws after an await, and when the failure report itself throws;
  restored exactly once; nothing propagates to the `async void` caller.
- `RecordingUiPlannerTests` — transcribing never sounds the start beep and never
  reads as recording; only recording does; idle leaves the delivered transcript
  in place.

## Documentation

- `troubleshooting.md` — two new entries: the delivery-refused message with its
  causes and remedies, and the unexpected-delivery-failure message.
- `dictation.md` — Mutation waits to see whether the text arrived before saying
  it worked.
- `keyboard-shortcuts.md` — a shortcut fires once per press; holding it does
  nothing extra.
- HTML regenerated with `Documentation\Build-UserGuide.cmd` and committed
  alongside.

## Not done

Verifying #226 against a physically held hotkey needs a real keyboard and a real
window — that one is for manual acceptance.

### DIFF
--- a/Documentation/UserGuide/html/dictation.html
+++ b/Documentation/UserGuide/html/dictation.html
@@ -229,11 +229,14 @@ or when you are dictating notes into Mutation itself and don't want stray text
 landing in another window.</p>
 <p>One thing worth knowing: if the Mutation window itself is the one in front, nothing
 is inserted anywhere. Mutation won't paste into itself.</p>
-<p>Whichever choice you make, Mutation waits to see whether the text actually arrived
-before it tells you it worked. If Windows refused the keystrokes — which it does
-when the app in front is running as an administrator — you get the failure beep and
-a message saying so, instead of a success beep for text that never landed. The
-transcript stays in the Mutation window either way, so nothing is lost.</p>
+<p>Whichever choice you make, Mutation waits for the typing or pasting to finish before
+it tells you the result, and checks that Windows accepted the keystrokes. If Windows
+turned them down, you get the failure beep and a message saying the text could not be
+sent, instead of a success beep for text that never landed. The transcript stays in
+the Mutation window either way, so nothing is lost.</p>
+<p>Windows does not always admit that it refused, so this catch is not a guarantee. If a
+success beep is ever followed by an empty window in the other app, the text is still
+in Mutation — paste it yourself with <strong>Ctrl+V</strong>.</p>
 <h3 id="sending-a-shortcut-afterwards">Sending a shortcut afterwards</h3>
 <p>In <strong>Settings</strong>, on the <strong>Speech to Text</strong> page, there is an optional box called
 <strong>Send hotkey after transcription</strong>. Whatever shortcut you put there is sent to the
--- a/Documentation/UserGuide/html/dictation.html
+++ b/Documentation/UserGuide/html/dictation.html
@@ -229,6 +229,11 @@ or when you are dictating notes into Mutation itself and don't want stray text
 landing in another window.</p>
 <p>One thing worth knowing: if the Mutation window itself is the one in front, nothing
 is inserted anywhere. Mutation won't paste into itself.</p>
+<p>Whichever choice you make, Mutation waits to see whether the text actually arrived
+before it tells you it worked. If Windows refused the keystrokes — which it does
+when the app in front is running as an administrator — you get the failure beep and
+a message saying so, instead of a success beep for text that never landed. The
+transcript stays in the Mutation window either way, so nothing is lost.</p>
 <h3 id="sending-a-shortcut-afterwards">Sending a shortcut afterwards</h3>
 <p>In <strong>Settings</strong>, on the <strong>Speech to Text</strong> page, there is an optional box called
 <strong>Send hotkey after transcription</strong>. Whatever shortcut you put there is sent to the
--- a/Documentation/UserGuide/html/keyboard-shortcuts.html
+++ b/Documentation/UserGuide/html/keyboard-shortcuts.html
@@ -157,6 +157,10 @@ listens.</p>
 <p>There is one exception, and it is noted in the tables below: <strong>Ctrl+Comma</strong> opens
 Settings, and that one only works when Mutation's own window is in front.</p>
 <p>Every default shortcut listed here can be changed. Nothing is fixed.</p>
+<p>A shortcut fires once per press. Holding the keys down a little too long does
+nothing extra — you will not start a second recording, or flip the mute back and
+forth, just because your fingers lingered. Let go and press again when you want it
+to happen a second time.</p>
 <h2 id="the-full-list">The full list</h2>
 <h3 id="microphone">Microphone</h3>
 <div class="table-wrap"><table>
--- a/Documentation/UserGuide/html/troubleshooting.html
+++ b/Documentation/UserGuide/html/troubleshooting.html
@@ -294,13 +294,16 @@ has its own message: &quot;The transcript could not be sent to the other applica
 be running with higher privileges than Mutation. It is available in the Mutation
 window.&quot; You get a failure beep with it, not the success beep.</p>
 <p>Windows will not let an ordinary program type into a window that is running with
-administrator privileges. It does not warn anybody — it simply throws the keystrokes
-away. Mutation now checks whether Windows accepted them and tells you when it did not,
-rather than announcing success for text that never arrived.</p>
+administrator privileges. It simply throws the keystrokes away. Mutation checks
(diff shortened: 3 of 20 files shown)


jfbosch:
## Independent review — findings and responses

An independent reviewer went over the diff with a correctness and concurrency lens.
Nine findings. Four are fixed in `fa91fe6`, two are filed as follow-ups, three are
answered below.

### Fixed

**A stuck state where the old code converged (their #1 and #6).** The most serious
finding, and a genuine regression introduced by this PR. Raising an explicit
`Transcribing` from the cancel branch pins a stale snapshot: the cancelled
transcription's `finally` posts its own `Idle` through the same dispatcher, and if that
`Idle` lands first the window settles on `Transcribing` with nothing left to correct
it — buttons disabled, transcript box read-only, for the rest of the session. Exactly
the symptom #234 exists to remove. The old flag-reading handler converged regardless of
delivery order; the snapshot does not.

Neither the cancel branch nor the `RecorderBusyException` handler changes any state, so
neither raises one now. The beep and the status message are their whole signal, and the
operation that owns the state announces its own `Idle` when it unwinds. The general
`catch` keeps `CurrentActivity` — safe there, because the busy case is the only one
with a concurrent owner and it is caught above.

**`GuardedUiOperation` let `restore` escape (their #4).** A throw from `restore` faulted
the task and propagated back through the `await` in the `async void` caller, straight
to `App.OnUnhandledException` — the exit the guard exists to close. Caught and routed to
`onReportFailed` now, with two tests: nothing propagates, and a restore that fails
part-way keeps what it managed first.

**`ButtonEnabled` was inert (their #7).** `UpdateSpeechButtonVisuals` honoured
`isEnabled` only in its `else` branch, so `RecordingUiPlan.ButtonEnabled` had no effect
for Record or Stop while `RecordingUiPlannerTests` asserted it — a green test over an
unchanged window. Every branch honours it. No behaviour change today; both values were
already `true`.

**Two over-claims in the guide (their #2 and #5).** `dictation.md` no longer promises
the check catches a refusal, and `troubleshooting.md` now tells the reader to look
before pasting again, because the SendKeys fallback can report failure for a paste that
actually landed.

### Filed as follow-ups

**#294 — `SendInput` may not detect UIPI at all.** The sharpest finding. #232 asked for
the return count to be checked, and it is; but Microsoft documents that `SendInput`
fails under UIPI without the return value or `GetLastError` indicating it. If that holds,
the elevated-foreground scenario #232 was filed about still announces success. The count
check catches genuinely refused input and the plumbing is right either way, so this PR is
a real improvement — but it may not close the headline case. #294 carries the manual
verification step and, if confirmed, integrity-level detection before injecting.

**#295 — "Transcribing..." is left in the box after a cancel.** Pre-existing; the old
handler did the same. Wants its own activity rather than a change to `Idle`, which must
keep leaving a delivered transcript alone.

### Answered, no change

**`MOD_NOREPEAT` now applies to router mappings too (their #3).** Their concern is that
the router is a general-purpose remapper, so holding a remapped key should repeat. It
does not work that way today: the router callback is
`SendHotkeyAfterDelay(to, 50ms)`, and `SendHotkeyViaSendInput` opens with
`WaitForModifierRelease(200ms)`. While the combination is held, nothing is sent at all —
every repeat queues up and fires as a burst the moment the user lets go. Hold-to-repeat
was already broken, and duplicated keystrokes on release is the worse failure.
`keyboard-shortcuts.md` also describes the router as a command remapper ("your video
call app mutes with Ctrl+Shift+M"), not a navigation aid. Applying the flag uniformly is
an improvement here, not a loss.

**The SendKeys fallback returns false even when it works (their #5, code half).**
Deliberate, and I would keep it. The fallback cannot confirm the target took the
keystrokes, and for a blind user an unnecessary warning is recoverable where a false
success is not — nothing later in the session contradicts it. The documentation gap they
identified alongside it was real and is fixed.

**Announcement now queues behind a possible `SendKeys.SendWait` (their #9).** Bounded
and low: `WaitForModifierRelease` ≤ 200 ms, `HotkeyChordDelayMs` 25 ms per chord,
`SendInput` synchronous, and `PostSendKeysAsync`'s task is discarded so `SendHotkey`
returns promptly. Real but small, and only on the fallback path.

### After the fixes

`dotnet test --configuration Release`: **1276 passed, 0 failed**. Guide HTML regenerated
and committed. The reviewer found nothing at security or data-loss severity — the
transcript is written into the window before any delivery attempt on every path, so no
failure mode in this diff loses the user's text.

PR #296, opened 2026-08-07
Delete code nothing calls, and stop five tests asking the clock

### LINKED ISSUE #253: [DEBT] Tests: replace wall-clock and file-mtime assertions with state assertions
## Context

Several tests assert on elapsed time or filesystem timestamps rather than on observable state.

### 1. mtime-based idempotence — `SettingsManagerMigrationTests.cs:95` and `:333`

```csharp
var beforeMtime = File.GetLastWriteTimeUtc(_tempPath);
manager.UpgradeSettings();
Assert.Equal(beforeMtime, File.GetLastWriteTimeUtc(_tempPath));
```

Windows file-time stamps advance with the system clock tick (~15.6 ms by default), and the write-then-upgrade sequence completes well inside one tick — so an unwanted rewrite is not reliably observable.

**Fix:** capture `File.ReadAllBytes(_tempPath)` before and after and `Assert.Equal` on the byte arrays. This is strictly stronger, since a rewrite would reformat the JSON to `Formatting.Indented`. The sibling `SettingsManagerPromptIdTests.cs:99-103` already does exactly this — follow that model.

### 2. Wall-clock **upper** bounds — `OcrServiceTests.cs:100` and `:249`

`Assert.True(stopwatch.Elapsed < TimeSpan.FromMilliseconds(40))` and `< 90 ms` for three back-to-back grants. A single GC pause or thread-pool starvation exceeds these budgets on a loaded CI agent.

**Fix:** assert observable state instead — `OcrService.GetSharedRequestWindowState().RequestsInWindow` before and after expiry already proves the window rolled over (`:240` does this correctly). The *lower*-bound assertions elsewhere in the file are fine; they only get more true under load.

### 3. Real sleeps as the only window for a wrong extra run — `DebouncerTests.cs:65` and `:106`

```csharp
Thread.Sleep(50);
Assert.Equal(0, Volatile.Read(ref runs));
```

The rest of the class is deterministic via the injected `ManualDelay`; these two negative assertions rely on 50 ms of real time being enough for a spuriously-scheduled continuation to land.

**Fix:** have `ManualDelay` expose the count of *completed* gates and assert the action ran exactly once per completed non-cancelled gate.

### 4. Over-specified exception type — `OcrServiceTests.cs:61`

`Assert.ThrowsAsync<TaskCanceledException>` holds only because cancellation lands inside `Task.Delay`. If the 50 ms `CancelAfter` fires while the limiter is between the `lock` and the delay, `token.ThrowIfCancellationRequested()` (`OcrService.cs:252`) throws a plain `OperationCanceledException`.

**Fix:** `Assert.ThrowsAnyAsync<OperationCanceledException>` — the same reasoning already applied at `OcrManagerTests.cs:262`.

### 5. Timing-based concurrency lower bounds — `OcrManagerTests.cs:573` and `:603`

`Assert.True(service.MaxConcurrency >= 2)` depends on two tasks overlapping within a 50 ms window. Deterministic today, but an implementation-detail dependency.

**Fix:** use an explicit rendezvous (`SemaphoreSlim`/`TaskCompletionSource` released once N callers arrive) so overlap is enforced rather than hoped for.

## Notes

Found during a full codebase audit. None of these fail today; all are load-dependent flake risk.

### LINKED ISSUE #257: [DEBT] Remove dead code: ComputerVision folder, unreferenced Core types, unused _playbackEpoch
## Context

Several unreferenced code paths, some carrying actively misleading intent.

### 1. `CognitiveSupport/ComputerVision/` — entire folder unreferenced

All six files have zero references outside the folder, including from tests: `ApiRateLimiter.cs`, `ComputerVisionOcrClient.cs`, `IDocumentOcrClient.cs`, `DocumentOcrJobInput.cs`, `DocumentOcrJobResult.cs`, `DocumentOcrJobStatus.cs`.

Worse than dead: `ApiRateLimiter.AcquireAsync` returns a `Releaser` whose `Dispose` **does nothing**, contradicting the class documentation that tells callers to wrap calls in `using`. It duplicates the live `OcrService.RequestRateLimiter`, which has 8 tests — so there are two throttling implementations and coverage on only one.

### 2. Unreferenced UI types

`Mutation.Ui/Core/OcrState.cs`, `Core/CaptureDeviceComboItem.cs`, `Core/ComboBoxItem.cs`, `Core/SpeechToTextServiceComboItem.cs`.

### 3. `TextToSpeechService._playbackEpoch` (`CognitiveSupport/TextToSpeechService.cs:74`)

Incremented at `:156, :215, :329, :374, :430` and **never read**. Its comment claims it exists "so a stale deferred task can detect it was superseded" — describing a supersession mechanism that does not exist. This matters because there is a real supersession race on those exact lines (tracked separately).

### 4. `ReadingProgress.HighestThresholdCrossed`

Referenced only by `Mutation.Tests/ReadingPositionAndProgressTests.cs`. Production now uses `ProgressWeavePlanner` + `SpokenWeaveMap.HighestMarkerPercentAtOrBefore`. Ten passing tests guard code nothing calls — misleading coverage.

### 5. Minor

- `OcrService.cs:119` — `imageBytes ?? Array.Empty<byte>()` on a definitely-assigned non-nullable `byte[]`.
- `OcrService.cs:323` — `RequestRateLimiter.Reset()` is private and never called.

## Requirements

- Delete 1–2 and 5. For 1, confirm nothing external depends on `IDocumentOcrClient` first; if the intent was to migrate OCR onto it, file that as its own story instead of leaving it half-landed.
- For 3, remove `_playbackEpoch` or implement it as part of the `Speak` supersession fix.
- For 4, delete both the member and its tests, or restore production usage.

## Notes

Found during a full codebase audit.

jfbosch:
Three small hygiene stories from the Ready column, batched because they
overlap in the same area (OCR service and its tests).

Closes #257
Closes #253
Closes #260

### Dead code removed (#257)

The unreferenced `CognitiveSupport/ComputerVision/` document-OCR path
(`ApiRateLimiter`, `ComputerVisionOcrClient`, `IDocumentOcrClient`, and the
three job types), three unused `Mutation.Ui/Core` combo/state types, and
`ReadingProgress.HighestThresholdCrossed`.

`ApiRateLimiter` was actively misleading rather than merely unused: its
`AcquireAsync` returned a `Releaser` whose `Dispose` did nothing, while the
class comment instructed callers to wrap API calls in a `using`. Anyone
adopting it would have got no throttling. The live, tested throttle is
`OcrService.RequestRateLimiter`; there is now one implementation, not two.

`HighestThresholdCrossed` was reached only by its own ten tests — production
announces progress through `ProgressWeavePlanner` and
`SpokenWeaveMap.HighestMarkerPercentAtOrBefore`. Coverage that guards nothing
reads as coverage that exists. `MaxThreshold` stays; the weave planner uses it.

`OcrReadingOrderSorter` lives in the same folder and is **not** removed — it is
called from `OcrService`.

### Flake-prone test assertions replaced (#253)

None of these fail today. Each holds only while the machine stays quick, so a
loaded CI agent could turn them red with no bug present.

| Test | Was | Now |
| --- | --- | --- |
| `SettingsManagerMigrationTests` x2 | file mtime unchanged | file **bytes** unchanged |
| `RequestRateLimiter_AllowsRequestsAfterWindowExpires` | elapsed < 40 ms | limiter window snapshot: saturated → empty → granted |
| `SharedRateLimiter_AllowsFreshBatchAfterWindow…` | elapsed < 90 ms | `RequestsInWindow` / `TotalRequestsGranted` |
| `RequestRateLimiter_HonorsCancellation` | `ThrowsAsync<TaskCanceledException>` | `ThrowsAnyAsync<OperationCanceledException>` |
| `DebouncerTests` x3 | `Thread.Sleep(50)` then assert | `ManualDelay` reports elapsed vs cancelled gates |
| `OcrManager` throttle tests x2 | `MaxConcurrency >= 2` inside a 50 ms delay | rendezvous forces the overlap; asserts **exactly** 2 |

The mtime assertions were the weakest of these: Windows file times advance only
on the ~15.6 ms system clock tick, and write-then-upgrade completes inside one
tick, so an unwanted rewrite was never reliably observable. Byte comparison is
strictly stronger — a rewrite would reindent the JSON. `SettingsManagerPromptIdTests`
already did it this way.

The rendezvous carries a 10 s timeout that records a flag, so a genuinely broken
throttle fails with a readable assertion instead of hanging the suite. Verified
by temporarily setting the rendezvous above the throttle limit: the test failed
with *\"Two OCR calls never ran at once\"* rather than deadlocking.

### graphify cache untracked (#260)

125 files, 2.8 MB, almost all generated AST cache keyed by the tool version.
`.gitignore` now covers `graphify-out/` and `git rm -r --cached` drops it from
the index; the working copy stays on disk so the tool keeps working locally.
Isolated in its own commit so the large deletion does not contaminate review.

## Already fixed in the tree

Two items in #257 needed no change and are called out in the commit rather than
silently skipped:

- `TextToSpeechService._playbackEpoch` is already gone.
- `RequestRateLimiter.Reset()` is **not** dead — the OCR tests invoke it
  reflectively to clear the shared static limiter between runs, and it carries a
  comment saying so.

## Testing

`dotnet test --configuration Release` — **1319 passed, 0 failed**. Targeted
reruns of the reworked tests x3, all green. The two rendezvous tests also got
faster, since they no longer wait out a real delay.

No user-visible behaviour changes, so no user-guide update is required.


jfbosch:
## Review by Claude

**Verdict: Approved**

Three hygiene stories batched into one PR: dead code deleted (#257),
load-dependent test assertions replaced with state assertions (#253), and the
generated graphify cache untracked (#260). All three do what their issues ask.
Five findings were fixed on the branch during review.

### Non-compliance
None. `dotnet build --configuration Release` is clean (0 warnings, 0 errors) and
`dotnet test --configuration Release` reports 1319 passed, 0 failed. No UI
changes, so no accessibility surface is touched. No hardcoded defaults, no
secrets. Tabs throughout, matching `AGENTS.md`.

### Story alignment
Good, with two deviations that are correctly documented rather than silently
skipped.

Issue #257 asks to delete `RequestRateLimiter.Reset()` as unused. It is **not**
unused — `OcrServiceTests` invokes it reflectively at five call sites to clear
the shared static limiter between runs, and the method already carries a comment
saying so. Deleting it would have broken five tests. Correctly left in place.
Likewise `_playbackEpoch` was already gone from `TextToSpeechService`, and the
redundant `imageBytes ?? Array.Empty<byte>()` no longer exists.

Issue #257 also describes `CognitiveSupport/ComputerVision/` as an entirely
unreferenced folder. It is not: `OcrReadingOrderSorter` lives there and is called
from `OcrService.cs:207`. The PR correctly deletes only the six dead files.

The requirement to file a follow-up if OCR was mid-migration onto
`IDocumentOcrClient` does not apply — no consumer exists anywhere and `OcrService`
calls `Azure.AI.Vision.ImageAnalysis` directly, so nothing was half-landed.

### Correctness
`ApiRateLimiter`'s deletion is the most valuable part of #257: `AcquireAsync`
returned a `Releaser` whose `Dispose` was empty while the class comment told
callers to wrap calls in a `using`. Anyone who had adopted it would have had no
throttling at all and no way to notice.

`ReadingProgress.MaxThreshold` survives and `ProgressWeavePlanner.cs:35` still
compiles against it; only `HighestThresholdCrossed`, which had no production
caller, is gone. The `Azure.AI.Vision.ImageAnalysis` package reference stays
needed by `OcrService`, so it is correctly left in the csproj.

Five issues were found and fixed (see **Fixes applied**). The two substantive
ones were in the replacement assertions themselves:

1. `OcrServiceTests.cs` — the new `RequestsInWindow == 1` and `== 3` checks read
   a *sliding* window immediately after a grant. They go stale the moment the
   window rolls, so a stall between the grant and the snapshot could fail them —
   reintroducing exactly the load dependency #253 exists to remove. Both were
   redundant anyway: the window is asserted empty immediately before, and a
   grant against an empty window cannot block by construction, so the cumulative
   `TotalRequestsGranted` is the complete and time-independent proof.
2. `OcrManagerTests.cs` — a timed-out rendezvous group left every subsequent
   caller to pay the full 10 s timeout as well, so a genuinely broken throttle
   would have taken 30 s or more to report. The first timeout now releases its
   gate and short-circuits the rest.

The `ManualDelay` gate accounting in `DebouncerTests` is sound. `Trigger()`
registers the gate synchronously (`RunAsync` runs inline up to its first await),
and `_cts.Cancel()` fires `TrySetCanceled` synchronously through the token
registration — so by the time the last `Trigger()` returns, the superseded gates
are already `IsCanceled`. The counts are settled at assertion time, not merely
current, which is what makes dropping the sleeps safe.

### Tests
This PR *is* test work, so the bar is whether the new assertions are strictly
stronger than what they replace. They are.

The mtime→bytes swap is the clearest win: the old assertion could not observe an
unwanted rewrite at all, because Windows file times advance only on the ~15.6 ms
clock tick and write-then-upgrade finishes inside one. Byte comparison also
catches a rewrite that reindents the JSON, which mtime never could.

The rendezvous converts `MaxConcurrency >= 2` from a hope into a requirement, so
the assertion tightens from at-least-2 to exactly-2. Its failure path was
verified by temporarily setting the rendezvous size above the throttle limit: the
test failed with the intended message rather than deadlocking, and after the fix
it does so in 10 s instead of 30 s. The reworked tests also run faster than the
originals, since they no longer wait out a real delay.

Ten tests were deleted alongside `HighestThresholdCrossed`. That is a reduction
in test count but not in coverage — they exercised a method with no production
caller, which reads as coverage that does not exist.

### Regressions
None found. Every deleted type was checked for references across `.cs` and
`.xaml` and had none outside its own file. The build is clean, so no consumer
was silently dropped. `.gitignore` now covers `graphify-out/` wholesale, and
`git rm --cached` left the working copy on disk, so graphify keeps working
locally with no rebuild.

### Minimality
Three stories in one PR is a deliberate batch, and the commits keep them
separable — the 125-file graphify deletion is isolated in its own commit exactly
as #260 asks, so it does not bury the code review. Nothing unrelated is bundled.

### Fixes applied
`b947220` — Review fixes: keep the clock out of the new assertions too

- Dropped the two window-count assertions that could go stale, keeping the
  time-independent `TotalRequestsGranted` (`OcrServiceTests.cs`).
- First rendezvous timeout now releases its gate and short-circuits later
  callers (`OcrManagerTests.cs`).
- Corrected the `ConcurrencyTrackingOcrService` class comment, which still
  credited the delay for producing the overlap.
- Documented that the expected call count must be a whole multiple of the
  rendezvous size.

No issues remaining.

---
Reviewed at b947220bfb36cb44e48292a878cc8089fc1b83eb
🤖 Reviewed with Claude Code

PR #387, opened 2026-08-16
Hold the speakers open, so a beep arrives with the thing it confirms

### LINKED ISSUE #386: [BUG] The success sound arrives seconds after the text it is confirming
## Context

After a dictation run finishes, the transcript is pasted into the application the user
was working in, and a success sound is meant to confirm it. In practice the sound often
arrives **5 to 10 seconds after the text has already landed**, and sometimes it arrives
straight away. The delay is not consistent from run to run.

For a user driving the app by ear, a confirmation that arrives ten seconds late is worse
than no confirmation at all. It is heard while they are already typing something else, so
it reads as a second operation finishing rather than as the one they were waiting for.

### What was measured

Two things were measured before writing this, so the cause is not guesswork.

**The app asks for the sound on time.** `Mutation.Hotkey.log` records both the paste
(`Sending via SendInput: 'Ctrl+V'`) and the shortcut sent immediately afterwards
(`SendHotkeyAfterDelay firing`), which is scheduled with a fixed 100 ms delay. The gap
between those two lines was measured across ~25 real dictation runs in the user's own log
and was 134–211 ms every single time — meaning the code between them took 34–111 ms, with
no outliers. `BeepPlayer.Play(BeepType.Success)` is the very next statement after that
scheduling call (`MainWindow.xaml.cs` `FinalizeTranscript`, and the two sibling delivery
sites). So the beep is requested well under a quarter of a second after the paste, every
time. The UI thread is not stalled, and nothing on the delivery path is slow.

**The request itself returns instantly.** `BeepPlayer` plays through
`System.Media.SoundPlayer`, which is Win32 `PlaySound` with `SND_ASYNC`. Timing
`SoundPlayer.Play()` on the user's own `Success.wav` gave 15 ms on the very first call and
0.3–0.5 ms after that — including while another thread was inside a `PlaySync()` loop.

So the sound is requested on time, the request returns immediately, and the sound is still
not heard for several seconds. The wait is inside Windows, after the app has handed the
sound over, and today the app has no control over it and no way to see it.

### Why Windows can be slow here

`PlaySound` owns no audio device. For each beep it opens the default render endpoint from
scratch on an internal thread, plays, and closes it again. Opening that endpoint is
normally a few milliseconds, but it is not bounded: it waits behind whatever else is
happening to the endpoint. Dictation is exactly the situation that disturbs it — the
recording has just closed a capture device, and on shared capture/render hardware
(a headset in particular) Windows reconfigures the endpoint when the microphone is
released. A beep asked for during that window waits for it.

There is a second fault in the same place, independent of the delay. `PlaySound` allows
one sound per process: starting a new one stops whatever is playing. Mutation's own retry
beeps (`BeepPlayer.PlayRepeated`, used by the transcription retry ladders) run a
`PlaySync()` loop on a background thread, so a retry beep and a success beep in the same
run cut each other off. Nothing coordinates them.

## Requirements

1. A beep must start within a few tens of milliseconds of the code asking for it,
   regardless of what the microphone was doing a moment earlier. Concretely: the success
   sound after a dictation must be heard as part of the same event as the paste, not
   seconds later.
2. Beeps must no longer cut each other off. A retry beep still playing when the success
   beep is asked for must not swallow it, and must not be swallowed by it.
3. No beep may block the thread that asked for it. `BeepPlayer.Play` and
   `BeepPlayer.PlayRepeated` stay non-blocking and safe to call from the UI thread.
4. The existing public behaviour is kept: custom `.wav` files when the user has turned
   them on, synthesized tones otherwise; the 500 ms duplicate-suppression window; the
   repeat counts used by the retry ladders; the settings-page preview; and the
   file-loading problems reported through `LastInitializationIssues`.
5. The latency is made visible, so if a sound is ever late again it can be proven rather
   than argued about.

## Notes

- The app already owns an audio output abstraction for playing recordings back —
  `IAudioOutputDevice` / `WaveOutAudioOutputDevice` over NAudio, with a test fake
  (`Mutation.Tests/FakeAudioOutputDevice.cs`). The beep path should use the same kind of
  device rather than WinMM's fire-and-forget `PlaySound`, so the device is opened once and
  kept open and the beep audio is already decoded in memory when it is needed.
- The user's custom beep files are a mix of 44.1 kHz and 48 kHz, all 16-bit stereo, so
  whatever plays them has to resample to one device format.
- Keeping a render device open has a cost worth stating: the app will show as an active
  playback session in the Windows volume mixer for as long as it runs, and on a Bluetooth
  headset it keeps the audio link awake. That is the trade being made for a confirmation
  sound that is actually on time.
- Call sites that must keep working: `MainWindow.xaml.cs` (three transcript delivery
  sites, the recording state change, the settings preview), `Services/OcrManager.cs`
  (screenshot and OCR outcomes), `AudioSessionManager.cs` (end-of-recording and failure),
  and `Extensions/ObjectExtensions.cs` (retry ladders).

jfbosch:
Fixes #386 — the success sound arriving seconds after the text it is confirming.

## What was wrong

After a dictation the transcript was pasted into the user's application and the success
sound followed five to ten seconds later, sometimes straight away, with no pattern to it.

## Where the delay actually was

Two measurements, before any code was written.

**The app asks for the sound on time.** `Mutation.Hotkey.log` records the paste
(`Sending via SendInput: 'Ctrl+V'`) and the post-operation shortcut scheduled a fixed
100 ms behind it. Across ~25 real dictation runs in the user's own log the gap between
those two lines was 134–211 ms — every time, no outliers. `BeepPlayer.Play(plan.Beep)` is
the statement immediately after that scheduling call, so the beep was requested inside a
quarter of a second of the paste on every run. The UI thread was not stalled.

**The request returns instantly.** Timing `SoundPlayer.Play()` on the user's own
`Success.wav`: 15 ms on the first call, 0.3–0.5 ms afterwards, including while another
thread was inside a `PlaySync()` loop.

Asked on time, returned at once, still not heard for seconds — so the wait was entirely
inside Windows, after the app had handed the sound over.

That is what `System.Media.SoundPlayer` is: Win32 `PlaySound`, which owns no audio device.
For every beep it opens the default output, reads, plays and closes again, on an internal
thread, with no bound on how long the open takes. The end of a dictation is the worst
moment to ask it: the recording has just released the capture device, and hardware that
shares a capture and a playback endpoint — a headset above all — is reconfigured by
Windows when the microphone is let go.

## What changed

`BeepAudioOutput` replaces `SoundPlayer`. It owns one audio device, opens it once and
holds it open, and feeds it from a `MixingSampleProvider` that beeps are dropped into.

- **Nothing is left to do at play time.** Beep audio is decoded and converted to the
  mixer's format (48 kHz stereo) when settings load — `BeepClipReader` handles the 44.1 kHz
  and 48 kHz files the user's own beeps come in, and the 22.05 kHz mono tones the app
  synthesizes for itself. Playing a beep is an array handed to a mixer.
- **No caller ever waits on an audio device.** Every request goes to one background thread.
  The callers are the UI thread finishing a dictation and Polly retry lambdas mid-
  transcription; neither can afford to block on a device or be thrown out of by one.
  `Play` returns immediately and cannot throw.
- **Sounds no longer cut each other off.** `PlaySound` allows one sound per process, so a
  retry beep still counting stopped the success beep that followed it and was stopped in
  turn by its own next repetition. They are added together now. The `PlaySync` loop, its
  `Task.Run`, and the generation counter that guarded it are all gone.
- **Failure is handled rather than assumed away.** A device that stops (unplugged,
  reconfigured, failed) is replaced on the next beep. One that will not open is reported
  once and retried after 30 seconds, instead of logging on every beep for the rest of the
  session.
- **The latency is now on the record.** Opening the device, or a beep waiting to reach the
  speaker, is logged whenever it exceeds 250 ms. If a sound is late again it can be
  measured instead of argued about.
- Device buffering is 100 ms across three buffers rather than NAudio's 300 ms default, so
  the app's own contribution to the delay is about 33 ms.

`BeepPlayer` keeps its whole public surface — `Initialize`, `Play`, `PlayRepeated`,
`PreviewFile`, `DisposePlayers`, `LastInitializationIssues`, the default sequences, the
500 ms duplicate-suppression window and the repeat counts the retry ladders rely on. No
call site changed.

## What the user will notice

The success sound arrives with the paste. Two consequences follow, both documented in the
guide:

- Mutation shows as an active playback session in the Windows volume mixer for as long as
  it runs, even when silent. Its level can be set there.
- On a Bluetooth headset the audio link is kept awake rather than dozing between sounds.

## Tests

`dotnet test --configuration Release` — **2465 passed, 0 failed.**

New: `BeepAudioOutputTests` (device opened once and held; `Play` returns in under 250 ms
against a device that takes a second to open; two beeps mix rather than truncate; a
stopped device is replaced; a broken device is reported once, retried after the backoff,
and never throws at the caller; shutdown releases the device), `BeepClipReaderTests`
(format conversion both ways, mono to stereo, whole frames only, overlong files trimmed,
non-WAV refused, plus the real shipped `.wav` files), `BeepClipSampleProviderTests`
(repeat counts survive reads that land mid-clip, an empty clip ends instead of spinning,
offset writes leave the rest of the buffer alone).

## What has not been verified

The fix removes the only remaining unexplained variable, and the mechanism is measured
rather than inferred — but the delay itself only appears on the user's own machine with
their own audio hardware, and an agent session cannot reproduce it. Confirming that the
success sound now lands with the paste needs a real dictation run on that machine. The
`[Beep]` log entries are there for exactly that.

### DIFF
--- a/CognitiveSupport/BeepAudioOutput.cs
+++ b/CognitiveSupport/BeepAudioOutput.cs
@@ -1,0 +1,320 @@
+using System;
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Threading;
+using NAudio.Wave;
+using NAudio.Wave.SampleProviders;
+
+namespace CognitiveSupport;
+
+/// <summary>
+/// The speaker every beep goes out of: one audio device, opened once and held open, fed by a
+/// mixer that beeps are dropped into.
+///
+/// <para>
+/// This replaces <c>System.Media.SoundPlayer</c>, and it exists because of what that API does
+/// between being asked for a sound and the sound being heard. It owns no device, so for every
+/// beep Windows opens the default output, reads the file, plays, and closes again — on an
+/// internal thread, with no bound on how long the open takes. A dictation run is the worst case
+/// for it: the recording has just released the microphone, and on hardware that shares a
+/// capture and a playback endpoint (a headset above all) Windows reconfigures the endpoint when
+/// the microphone is let go. A beep asked for during that window waits it out. Measured on the
+/// user's machine, the app was asking for the success beep 40 to 110 ms after the transcript was
+/// pasted, every time, and the sound was still arriving five to ten seconds later (issue #386).
+/// </para>
+///
+/// <para>
+/// Holding the device open moves that cost to startup, where nobody is waiting on it, and leaves
+/// a beep needing nothing but an array handed to a mixer. It also ends the one-sound-at-a-time
+/// rule that came with the old API: sounds are added together now, so a retry beep still
+/// counting cannot swallow the success beep that follows it.
+/// </para>
+///
+/// <para>
+/// The trade is worth stating plainly. Mutation now shows as an active playback session in the
+/// Windows volume mixer for as long as it runs, and on a Bluetooth headset it keeps the audio
+/// link awake. That is the price of a confirmation sound that arrives with the thing it is
+/// confirming.
+/// </para>
+///
+/// <para>
+/// Every request is handed to one background thread rather than played on the caller's. Opening
+/// a device is the slow step and it can fail, and the callers here are the UI thread finishing a
+/// dictation and a Polly retry lambda mid-transcription — neither can afford to wait on an audio
+/// device or to be thrown out of by one.
+/// </para>
+/// </summary>
+public sealed class BeepAudioOutput : IDisposable
+{
+	/// <summary>The one format the mixer runs at; every clip is converted to it when loaded.</summary>
+	public const int SampleRate = 48000;
+
+	public const int Channels = 2;
+
+	/// <summary>
+	/// How much audio the device buffers, and in how many pieces. A beep starts at the next
+	/// buffer boundary, so this is the app's own share of the delay between asking for a sound
+	/// and hearing it: about 33 ms here. NAudio's default is 300 ms in two buffers, which would
+	/// have put a tenth of a second between the paste and its confirmation for no reason.
+	/// </summary>
+	public const int DesiredLatencyMs = 100;
+
+	public const int BufferCount = 3;
+
+	/// <summary>
+	/// Anything slower than this is worth a line in the log. The point is that the next time a
+	/// beep is late, whether the app was the cause is a matter of record rather than argument
+	/// (issue #386).
+	/// </summary>
+	public const int SlowReportThresholdMs = 250;
+
+	/// <summary>
+	/// How long to leave a device that would not open alone before trying again. Without it, a
+	/// machine with no working audio output logs a failure for every beep of the session.
+	/// </summary>
+	public static readonly TimeSpan ReopenBackoff = TimeSpan.FromSeconds(30);
+
+	private readonly Func<IAudioOutputDevice> _deviceFactory;
+	private readonly Action<string, string> _log;
+	private readonly Func<DateTimeOffset> _now;
+	private readonly BlockingCollection<PlayRequest> _requests = new();
+	private readonly Thread _pump;
+
+	// Raised whenever the queue empties. The pump is asynchronous by design, so a test needs a
+	// way to say "and now everything asked for has been handed over" without sleeping. The count
+	// and the event are moved together under one lock, or a beep queued in the instant the pump
+	// finished the one before it would leave the event saying there was nothing left to do.
+	private readonly ManualResetEventSlim _idle = new(initialState: true);
+	private readonly object _idleLock = new();
+	private int _pending;
+
+	private IAudioOutputDevice? _device;
+	private MixingSampleProvider? _mixer;
+	private DateTimeOffset _nextOpenAttempt = DateTimeOffset.MinValue;
+	private volatile bool _deviceFaulted;
+	private bool _disposed;
+
+	public BeepAudioOutput()
+		: this(() => new WaveOutAudioOutputDevice(DesiredLatencyMs, BufferCount))
+	{
+	}
+
+	/// <param name="deviceFactory">Makes the output device. Injected so the rules here can be tested without audio hardware.</param>
+	/// <param name="log">Where slow opens and failures are reported. Defaults to the app's error log.</param>
+	/// <param name="now">The clock, so the reopen backoff can be tested without waiting on it.</param>
+	public BeepAudioOutput(
+		Func<IAudioOutputDevice> deviceFactory,
+		Action<string, string>? log = null,
+		Func<DateTimeOffset>? now = null)
+	{
+		_deviceFactory = deviceFactory ?? throw new ArgumentNullException(nameof(deviceFactory));
+		_log = log ?? ErrorLogger.LogInfo;
+		_now = now ?? (() => DateTimeOffset.UtcNow);
+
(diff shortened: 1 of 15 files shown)


jfbosch:
## Independent review, and what it changed

Two reviewers ran against this: a code review of the diff, and a verification pass that
checked every assumption the design makes about NAudio 2.3.0 and Windows audio against
NAudio's own source at tag `v2.3.0` and against the shipped binary.

The code review found no defect at its confidence bar, and confirmed the parts that
mattered most: NAudio's mixer really does lock the same list for `AddMixerInput` and
`Read`, so feeding it from the pump thread while the device reads is safe; `PlaybackStopped`
really does arrive on NAudio's own thread here, and the only state it touches is one
volatile flag; `BeepPlayer.SyncLock` never covers anything that blocks on the beep path.

The verification pass **refuted one assumption** and turned up two risks. Four changes came
out of it, in `31d7031`:

**Shutdown could have ended the process.** If the beep thread was still inside a device
open when the window closed, `Dispose` gave up after two seconds and disposed the queue and
the event underneath it. The thread then threw out of its `finally`, and an exception
escaping a background thread's `finally` takes the process with it. It now only tidies up
behind a pump that actually finished, and the `finally` is guarded. This is not a
hypothetical path — a slow device open is the exact failure this whole change is about.
`BeepAudioOutputShutdownTests` covers it, watching for an unhandled exception.

**Beep files that used to play would have gone quiet.** NAudio refuses two shapes
`PlaySound` handled without complaint: the `WAVE_FORMAT_EXTENSIBLE` layout, which many
tools write for float audio and which Windows requires above two channels, and compressed
formats such as ADPCM and mu-law. A file the user had been hearing for months could have
stopped working the day this shipped. Extensible files are now relabelled from the
sub-format GUID in their header, and compressed ones go through Windows' own codecs — the
same ones the old path relied on. Both extensible variants are tested.

**The mixer is now checked before it is fed.** NAudio adds an input to its list *first* and
validates the format *afterwards*, outside the lock — so a wrong-format clip is already in
the mix by the time the check throws, and would then play at the wrong rate over everything
else. The format is checked before the handover instead.

**The refuted assumption, written down.** The mixer drops an input on *any* short read, not
only on a zero one — the test is `samplesRead < count`. The samples in that short read are
mixed before the input is dropped, so nothing is lost when a beep ends this way. But it
means a beep must fill every read completely until it genuinely finishes. It already does;
the requirement is now stated on the provider so nobody later turns it into a truncation
bug.

## Two things the review could not verify, now verified

- **Message parity.** `LastInitializationIssues` produces the same six strings as before,
  character for character (`Could not load {start|success|failure|end|mute|unmute} beep
  file: {path}`), and `PreviewFile` keeps its signature and its true/false contract.
  Checked against `origin/main`.
- **What `Initialize` costs on the UI thread.** Measured: 29 ms to decode all six shipped
  beep files, holding 3.2 MB of audio. That is what makes doing the work there — rather
  than at the moment a beep is wanted — the right trade. There is a budget test on it now,
  with wide enough headroom not to flake on a loaded machine.

`dotnet test --configuration Release` — **2470 passed, 0 failed.**

